PR #8, opened 2019-02-07
reenable inline functions on windows by updating bindgen

### LINKED ISSUE #7: Reenable the binding generation of inline functions under Windows.
For fixing the build under windows, the inline function binding generation (e9efde52a250f1d377c495498f96e239a4c60856) was disabled and caused presumably useful static functions to be removed from the bindings.

The problem has been fixed in bindgen recently (https://github.com/rust-lang/rust-bindgen/pull/1513) (I've tested that with a local clone). As soon a new version containing this fix is released, the binding generation of inline function should be reenabled.

pragmatrix:
Fixes #7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,4 +10,5 @@ links = "skia"
 
 [build-dependencies]
 cc = "1"
-bindgen = "0.46"
+# temporary, 47.2 will contain the changes that are required to support inline functions on windows.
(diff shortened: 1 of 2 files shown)


pragmatrix:
correction, first attempt was a bad gateway (502) on this URL: https://chromium.googlesource.com/chromium/deps/icu.git/
interesting.

Brooooooklyn:
maybe it caused by https://github.com/Brooooooklyn/rust-skia/blob/master/build.rs#L31

PR #15, opened 2019-02-27
Preliminary Vulkan support and more extern C functions for skia-safe.

pragmatrix:
This brings my fork on par with the one in the rust-skia organization and contains full support for building skia-safe.


pragmatrix:
Note that this PR also removes the hello / canvas example. It has been moved and ported to skia-safe. If this is not the intention, we can revive it here.

PR #16, opened 2019-02-28
Canvas bindings

pragmatrix:
This PR contains more bindings and stays in WIP until rust-skia/skia-safe#3 is complete.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,5 @@ vulkan = []
 
 [build-dependencies]
 cc = "1"
-# temporary, 47.2 will contain the changes that are required to support inline functions on windows.
-bindgen = { git = "https://github.com/rust-lang/rust-bindgen", rev = "49b74244f7db8618b9a1b135ce2659751636c59c" } 
+bindgen = "0.47.3"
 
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -14,6 +14,5 @@ vulkan = []
 
 [build-dependencies]
 cc = "1"
-# temporary, 47.2 will contain the changes that are required to support inline functions on windows.
-bindgen = { git = "https://github.com/rust-lang/rust-bindgen", rev = "49b74244f7db8618b9a1b135ce2659751636c59c" } 
+bindgen = "0.47.3"
```
Brooooooklyn: `0.48.1` available, could this version fix some `inlining` issues?

pragmatrix: The basic support for linking functions marked `inline` was fixed already in 0.47.2, but there are a number more linker errors, caused by small functions defined in the header file (inline or not) that call out to others. I've marked them in code with "does not link" comments and replaced them with either an `C_` function or a custom implementation on the Rust side if it was simple enough.

Presumably, this is a bindgen name mangling problem, or the compiler just does not compile them into the lib file. Someone needs to investigate that.

Funny thing: I could predict with high accurracy when a function would cause a linker error, but never found an actual pattern.

Regarding `0.48.1`, there were two struct layout fixes that could be interesting, but I think we should postpone that update until we have the projects combined and the CI set up.

pragmatrix:
both PRs are ready from my side. If nothing comes up, I will merge them tomorrow.

pragmatrix:
Thanks!

Issue #6, opened 2019-02-03
Suggestion: Distribute the raw bindings as a crate and continue to use them as dependency.


pragmatrix:
I did an experiment and separated the canvas implementation and the `hello.rs` into a another dependent project and it went along fine, Rust seems to be awesome in this regard, the `skia.lib` and the `skiabinding.lib` were all included into the resulting rlib, and the linker flags were propagated.

The improvements I seek are:

- To separate the the evolution of the raw binding generation (including feature control!) from the future _safe_ bindings or other _simplified_ bindings like the already existing canvas implementation.
- To simplify IDE support and development. For example, CLion completely freaks out with the included raw bindings inside the same project, it shows errors and the IDE completion does not work, but as soon they are in a dependent create, everything is fine.
- To reduce build times for creating dependent crates, running `build.rs` for every change seems to be too heavy.
- To simplify builds. We could run the binding generation every time (i.e remove the `INIT_SKIA` switch). Cargo seems to be intelligent enough to build the project only once for dependent projects as long there are no changes affecting it, though I don't know how reliable that is.
- To avoid waiting for and using azure-pipelines for uncountable hours and heating up our planet even more.

Note that I am a Rust beginner, and some of the points above may be speculative and based on naive assumptions.

Naming suggestions for the resulting crates and projects: 

- `skia-raw`, for the raw bindings (we just export everything that `binding.rs` produces).
- `skia-canvas`, for the simplified canvas drawing implementation.
- `skia-safe`, for the safe bindings which should closely resemble the Skia API.

Brooooooklyn:
I approve your suggestion. More over it, I'm thinking about make ***skia-raw*** a prebuilt library which include `skia static link file which generated in CI. Build skia is painful if someone want to use `skia-raw`.

pragmatrix:
> More over it, I'm thinking about make _**skia-raw**_ a prebuilt library

I would like that but have no idea how a built a crate with prebuilt libraries for multiple platforms. Do we need to split them up, in `skia-raw-{linux,windows,macos}`, and then refer them from `skia-raw` with feature flags?

Some progress on my side: I've built something similar to what I would consider a safe Skia API on top of the bindings and ported the canvas drawing example to it. It's a mess, but as soon I do have some time, I will clean it up and try to formulate how I think we can proceed.

pragmatrix:
I think we can build on that: 

https://github.com/pragmatrix/skia-safe

[...]

PR #18, opened 2019-03-09
Merge skia safe

pragmatrix:
Merges skia safe into the repository and also runs its testcases on the azure pipelines.

Related to #6.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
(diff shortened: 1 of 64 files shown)


pragmatrix:
@Brooooooklyn I've tried updating bindgen to 0.48.1 to fix the linker errors on Linux, but no luck, also there are significantly more bindgen layout problems on Linux, but so far it affects only reference counted types, which should not be a problem for skia-safe right now.

pragmatrix:
@Brooooooklyn fyi I've added my name to the LICENSE file.

PR #30, opened 2019-03-10
[WIP] [EXPERIMENTAL] fix a build problem on macosx

pragmatrix:
I can't build Skia on one of my MacOSX machines, the error appear to be the same as in https://github.com/aseprite/aseprite/issues/1981, so let's take a look if that fix does not destroy the CI builds on azure.


pragmatrix:
That "fix" caused a linker error in the end, installing the XCode Command via Xcode > Open Developer Tool > More Developer Tools... fixed the problem.

PR #44, opened 2019-03-13
Skia org example artifacts

pragmatrix:
This PR tries to generate the PNG images of the skia-org example on azure and stores them into a downloadable build artifact. 

This way we can automatically generate the example images on all platforms.


pragmatrix:
Artifacts may be downloaded from the azure pipelines page: https://dev.azure.com/pragmatrix-github/rust-skia/_build/results?buildId=270 (button top right) or the summary tab.

I'm not completely satisfied, because I now think that it would be better to combine them all into one archive, or perhaps even deploy them to a image hosting provider.

But for now, this is sufficient.

PR #45, opened 2019-03-14
Safe API wrappers for realizing all drawings in skia.org's SkPaint overview

pragmatrix:
https://skia.org/user/api/skpaint_overview

- [x] SkTextBlob
- [x] SkShader
- [x] SkGradientShader
- [x] SkImage::makeShader()
- [x] SkPerlinNoiseShader
- [x] SkTableColorFilter
- [x] SkPath2DPathEffect
- [x] SkLine2DPathEffect
- [x] SkPath1DPathEffect
- [x] SkCornerPathEffect
- [x] SkDashPathEffect
- [x] SkDiscretePathEffect
- [x] Extend the skia-org example.


pragmatrix:
The TableColorFilter example is broken, I assume that the endianess detection for the N32 mode of the Canvas has a problem (we had another related testcase failing before). I'm waiting for the results of the other platforms to confirm that.

And the last three examples use random values and look different compared to the website, which is  concerning, because path effects use a custom platform independent random number generator.

pragmatrix:
My assumption was wrong, the pointer computation for an sized array ref inside an option was wrong. It's Fixed now.

pragmatrix:
Some highlights:

01-three-paints

08-sweep-gradient-shader

10-turbulence-perlin-noise-shader

19-dash-path-effect

pragmatrix:
The examples that use random values look the same on all platforms. So I guess we can leave it at that.

Brooooooklyn:
nice job!

Issue #21, opened 2019-03-04
Better Enum representation.


pragmatrix:
There are some situations for which it would be convenient to pass a mutable pointer to an enum handle to a native function.

PR #46, opened 2019-03-16
Use the actual Matrix subscript operator instead of accessing the mat…

pragmatrix:
…rix's values directly

This PR fixes a bug that causes the Matrix's type not to be recomputed when matrix values (like perspective for example) were changed.


pragmatrix:
The test fails because the `enum SkMatrix_TypeMask` returned from `SkMatrix::getType()` is defined as a Rust Enum and the value `0xf` is not a valid value of that Enum (even though casted to an `u32` right after retrieving it from the C function).

I've taken a look at the assembler code and the compiler _does not even create code for the comparison with `0xf`_ in release builds, because it assumes that the value can _never_ be `0xf`, even though it was casted to an `u32` before.

This has several consequences:

_rustified_ Enums can not be used for Flags. This one is quite obvious and an oversight on my side.

I need to review all Enums that are returned from the C functions, if one of them _might_ return a value not covered in the Enum, we can't use a Rustified Enum there.

I am thinking about going back to bindgen's `constified` Enums by default. Originally, I've enabled rustified Enums just for convenience, because I disliked these very long identifier bindgen produces for the constified Enums. 

In addition to that we may need to verify all enum values that are retrieved from Skia against ours, at least in debug builds, but I've yet have to find a decent crate that does that. Probably it's time to write a derive macro that provides a simple validation function. As soon [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) lands in stable, a `Enum::try_from(u32) -> Option<Enum>` should be good enough (see also #21).

PR #47, opened 2019-03-17
Use bindgen "constified enums" for all bitflag like enums.

pragmatrix:
See here why: https://github.com/rust-skia/rust-skia/pull/46#issuecomment-473657661

This branch is based on #46.


pragmatrix:
Thank you for looking into that.

PR #52, opened 2019-03-18
Integrate depot tools

### LINKED ISSUE #48: Integrate depot_tools
I think we should integrate `depot_tools` into the project, probably as a submodule. 

This way, we can run `ninja` from `build.rs` and it will be a bit easier to compile this library. 

In addition to that we can pin down the exact version of the `depot_tools` that we support, currently we are pulling the master branch for the CI builds.

Interesting enough, I've never installed `depot_tools` on my machine, so it should actually be enough to pull in `ninja` only.

/cc @Brooooooklyn
Related to #3

pragmatrix:
First attempt to ~pick up~run ninja directly from a clone of `depot_tools` that's a submodule in the `skia-bindings` directory.

Closes #48

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "skia"]
 	path = skia-bindings/skia
 	url = https://github.com/google/skia.git
+[submodule "skia-bindings/depot_tools"]
+	path = skia-bindings/depot_tools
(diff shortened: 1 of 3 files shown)


pragmatrix:
It worked! No ninja or depot_tools installation needed anymore! :tada:.

Brooooooklyn:
awesome!

PR #56, opened 2019-03-18
GPU Support

### LINKED ISSUE #50: Render a second set of the skia-org examples with OpenGL.
For that I'd like to complete the SkSurface bindings and most of the GPU support in `include/gpu`.

pragmatrix:
This PR has the following objective:

- [x] review and modernize the Vulkan feature APIs.
- [x] support OpenGL
- [x] support most of the GPU related SkSurface functions.
- [x] extend GrContext with OpenGL support functions.
- [x] render the skia-org example with OpenGL

Closes #50


pragmatrix:
Sad, the examples can not be rendered with OpenGL when there is no possibility to open a (dummy) window. Locally it works under Windows (Radeon R7 200) MacOSX (Intel HD 3000) just fine, although the 5th example of the Path examples does not render a zero sized square cap on both machines, but this should be investigated separately.

Under Linux with XMing 6.9 on the Windows side (7 is available, but behind a donor gate, and the changelog does not tell if XInput2 is supported):
```
thread 'main' panicked at 'X server has XInput extension 1.4 but does not support XInput2'
```
So I assume this would work under Linux with a XInput2 capable X server, too. 

I'll try to get Vulkan rendering to work in separate PR, and remove the OpenGL artifact generation from the CI as long we can't find a way to render OpenGL without a Window.

Brooooooklyn:
In my experience, render OpenGL without a Window was simple under the macOS, but macOS not support OpenGL anymore since OpenGL 4.1.
Under linux, it seems impossible to create OpenGL context without any window. I see most other application choose `Mesa` to mock OpenGL API offscreen.
you can see: https://github.com/servo/rust-offscreen-rendering-context

pragmatrix:
@Brooooooklyn Thanks for that, the offscreen context works better on Windows where the mouse pointer does not reappear, but it attempts to Open an X Window under Linux.

I am preparing a PR to see how it goes works on MacOSX on Azure. Even if we can't use Azure for rendering OpenGL, it seems to be a better choice already.

PR #62, opened 2019-03-26
Use offscreen gl context to render the skia-org examples

pragmatrix:
Following up [from here](https://github.com/rust-skia/rust-skia/pull/56#issuecomment-476454459), we may have better chances to render the OpenGL examples by using https://github.com/servo/rust-offscreen-rendering-context instead of glutin.


pragmatrix:
Interesting, so on Azure there _is_ a GPU in the Mac OS X instance. 

Anyway, I've removed the azure generation for OpenGL and think we should keep using the offscreen gl context.

PR #63, opened 2019-03-27
Support to render skia-org examples via ash on Vulkan devices.

### LINKED ISSUE #60: Support --driver vulkan in the skia.org example when the feature "vulkan" is enabled.
The necessary preparations are in #56.

pragmatrix:
Closes #60


pragmatrix:
Finally, this works (mostly).

Test results:
- Two of my Windows machines, one with an NVidia card and the other with AMD Radeon rendered the examples just fine. Another one with an Intel card crashes in the device creation function, but other Vulkan demos, do, too, so I suspect that this is a driver problem.
- For my Linux box, I did not found a suitable driver for the integrated graphics card.
- An older 2011 10.13 Mac had no Metal support, so Vulkan was out of the question.
- On a current 10.14 Mojave machine with Metal support, installing the VulkanSDK with MoltenVK worked fine, although I found the installation a bit confusing.

So as long the CI states otherwise, I think this can be merged.

PR #70, opened 2019-04-06
Fix build errors of the skia-org example when the feature vulkan is enabled

pragmatrix:


pragmatrix:
@Brooooooklyn  thanks! I wouldn't have bothered you with this one though.

PR #76, opened 2019-04-12
SkDocument & PDF Support

pragmatrix:
I am trying to figure out how to properly support `SkDocument` & PDF. 

In `SkDocument` the `SkCanvas` instances are bound by the current page's lifetime and so it seems that we need to wrap this differently to guarantee Rust's safety properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ to generate some Skia drawn PNG images in the directory `OUTPUT_DIR`. To render 
 
 `cargo run --example skia-org -- [OUTPUT_DIR] --driver opengl`
 
+And `cargo run --example skia-org -- --help` shows the drivers that are currently supported.
+
 ## Status
 
 ### Crate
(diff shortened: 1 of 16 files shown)


pragmatrix:
For a safe abstraction, I think that something in the realm of typestate programming is needed: https://yoric.github.io/post/rust-typestate/.

pragmatrix:
This is looking good. Build artifacts contain all the examples rendered in PDF: https://dev.azure.com/pragmatrix-github/rust-skia/_build/results?buildId=612&view=results

For example "Hello Skia" created on Linux:

[hello-skia.pdf](https://github.com/rust-skia/rust-skia/files/3078667/hello-skia.pdf)

pragmatrix:
I've removed XPS from the list of features we support, because it seems to be very Windows specific and the SkXPSDocument _requires_ a COM component to produce XPS.

PR #82, opened 2019-04-16
SVG Support

pragmatrix:
This PR is based on the PR in #76 and implements a wrapper for `SkSVGCanvas::Make()` to support SVG. 

SVG will be an optional feature that can be enabled in `Cargo.toml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ We do support most of the SkCanvas, SkPaint, and SkPath and related APIs and are
 - [x] Basic Drawing: Surface, Canvas, Paint, Path.
 - [x] Basic Effects and Shaders.
 - [x] PDF
-- [ ] SVG
+- [x] SVG
 - [ ] Animation
 - [x] Vulkan
 - [x] OpenGL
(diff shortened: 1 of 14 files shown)


pragmatrix:
SVG support is currently an experiemental in Skia itself, but basic drawing seems to work quite well so far.

pragmatrix:
This is ready to review, rendered SVG examples can be downloaded with the build artifacts: https://dev.azure.com/pragmatrix-github/rust-skia/_build/results?buildId=636&view=results

pragmatrix:
All back, [WIP] again, I will try to support reading and rendering of SVG with this PR to make SVG support complete.

pragmatrix:
Well, rendering of SVG is behind a experimental library that is only compiled when Skia is built with all its tools enabled adding up to around 2000 build steps instead of the usual 800-900, so I think we'll have to postpone until it gets out of the experimental stage, or we find some other way to render SVG (like SkiaSharp does).

pragmatrix:
Rebased on master.

Issue #103, opened 2019-05-08
Building on MacOS with libjpeg-turbo compile errors.


therealbnut:
Hey,

I had some compile errors related to `libjpeg` (you can see other people getting the same errors here: https://github.com/aseprite/aseprite/issues/1981).

I tried `brew install libjpeg-turbo` but that didn't work, in the end I had to put  `("skia_use_libjpeg_turbo", no()),` into `build.rs` to get it building.

I'm not sure if that's a change you want to include unconditionally, but I thought I should mention it in case someone else has the same issue.

Brooooooklyn:
Have you tried to install `/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg` ?

therealbnut:
I did, but I might need to re-verify again.

I also had to fix some issues with my python install so I re-installed my Xcode CLI tools. I’ll re-verify tomorrow without the build.rs changes, just in case it’s no longer needed.

therealbnut:
That works thanks :D

For reference this was what happened:
 * Try to build, but had issues where it couldn't find system `python` (I had a brew version instead)
 * Run `macOS_SDK_headers_for_macOS_10.14.pkg`, but that didn't fix it
 * Fixed `python` by reinstalling the system python by clean installing Xcode CLI tools (this removed `TargetConditionals.h`, but I didn't realise)
 * Run into `libjpeg-turbo` issue
 * Fix `libjpeg-turbo` by modifying `build.rs`
 * Have `TargetConditionals.h` issue for the first time
 * Run `macOS_SDK_headers_for_macOS_10.14.pkg` again (which removed the need for the `build.rs` change but I didn't realise)

pragmatrix:
Thanks for the detailed feedback! I will try to reproduce that and update the readme.

PR #68, opened 2019-04-06
M74

pragmatrix:
Because Chrome 74 ~~is around the corner~~may be out soon, we should be prepared for the release. 

This PR contains the needed changes for updating rust-skia to the Skia branch [chrome/m74](https://github.com/google/skia/tree/chrome/m74) (which might be updated while this PR is open!).

- [x] Make skia-safe compilable
- [x] Test skia-org examples.
- [x] Do the bindgen generated layouts fit?
- [x] Add new functions in existing wrappers.
- [ ] Add new include files (pretty much optional)
  - [ ] SkContourMeasure
  - [ ] SkCubicMap
- [x] Update README.md ([Rendered](https://github.com/rust-skia/rust-skia/blob/m74/README.md))
- [x] Update the versions of the `skia-bindings` and `skia-safe`.

PR #105, opened 2019-05-09
note that libjpeg-turbo may also fail to compile on macOS when the SD…

pragmatrix:
…K headers were not properly installed

Details are in this comment: https://github.com/rust-skia/rust-skia/issues/103#issuecomment-490737066

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,11 @@ This project attempts to provide _up to date_ safe bindings that bridge idiomati
 
 Note that the information in this section is preliminary. Please open an issue for any build problem.
 
-This project requires LLVM, python, and git to build.
+This project requires LLVM, Python 2, and git to build.
 
 To test if LLVM is installed with the correct version, use `clang --version`. Currently, version 7.0.1 is required, or - on macOS X - Apple LLVM Version 10 should do, too.
 
-For python, at least version 2.7 should be available. Use `python --version` to see what's there.
+For Python, at least version 2.7 _should_ be available! Use `python --version` to see what's there.
 
 ### macOS X
 
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ For python, at least version 2.7 should be available. Use `python --version` to 
 ### macOS X
 
 - Install the XCode command line tools with `xcode-select --install`.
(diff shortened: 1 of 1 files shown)


therealbnut:
It took some searching to know it was `libjpeg-turbo`, you might want to include reference to `SkJpegUtility.cpp` in a similar way that you referred to `TargetConditionals.h`

therealbnut:
This is the post where I found out it was libjpeg-turbo, perhaps using those error messages might be more representative:
https://groups.google.com/forum/#!msg/skia-discuss/OkMe3HWoV-U/Hxs872ySAgAJ

pragmatrix:
I wasn't able to reproduce the build problem of `SkJpegUtility.cpp`, but had the same error once when I integrated the M74 branch (#68).

pragmatrix:
[Rendered](https://github.com/rust-skia/rust-skia/blob/macos-libjpeg-turbo-readme/README.md)

therealbnut:
I think Skia specifically searches for the system python, and needs the symlinked dir `/usr/local/opt/python2` iirc. I believe I had 2.7 via brew in my `$PATH`, but Skia didn't find that. Perhaps you need to remove the system's symlink to replicate the issue I had.

Although IMO you've already gone above and beyond, thank you :) I think these issues are mainly with `skia`, not `rust-skia`.

pragmatrix:
There is something wrong with the azure pipelines, unlikely to be related to the changes. I am merging this. @therealbnut Thank you for taking a look at the changes, I would very much be interested in your further endeavor to use Skia with Rust.

therealbnut:
@pragmatrix thanks, at the moment we're just doing an exploration of various libraries and it looks pretty promising, I'll let you know if we end up using it, or why we didn't if we don't.

Out of interest, do you know any big projects using it? what's driving the development at the moment?

pragmatrix:
Thank you, tell us if there is something missing from it that would block your plans. 

I don't know of _any_ projects using it yet. My personal motivation is to get a deeper grasp of Rust and in the long term to create a development environment for native applications in which all possible states can be simulated _and tested_ anytime, in every detail, and in non-real-time. @Brooooooklyn and @M-Adoo seem to be interested in a WebAssembly port, of which I am not aware of the details.

PR #112, opened 2019-05-13
Build crate packages on Azure and deliver them to rust-skia.

### LINKED ISSUE #110: Build, verify, and deploy cargo crates of skia-bindings and skia-safe to rust-skia.
Now that #97 has landed, the next step to prepare for official crate releases is to build and verify the crates on the CI server and release them to the ~~skia-binaries~~rust-skia repository. The resulting `.crate` files are platform and feature independent, so we can create and verify them without an initial build in a separate Azure Pipelines job, but I would rather create them _after_ a binaries package was created so that the verification _already_ pulls in the previously built binary from the regular build and indirectly so tests the binary package downloader with each CI build. 

For a first setup, I think we should limit the crate generation to "Linux stable" for now.

pragmatrix:
Closes #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ Please share your build experience so that we can try to automate the build and 
 
 For situations in which Skia does not build or needs to be configured differently, we support some customization support in `skia-bindings/build.rs`. For more details about how to customize Skia builds, take a look at the [README of the skia-bindings package](skia-bindings/README.md).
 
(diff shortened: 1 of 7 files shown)


pragmatrix:
I think now that it's fine to deliver them to rust-skia, as long as they are marked as pre-releases.

pragmatrix:
Crates are looking good: https://github.com/rust-skia/rust-skia/releases/tag/0.10.1

Brooooooklyn:
Looks great

pragmatrix:
@Brooooooklyn Thanks! I've updated the readme accordingly.

PR #114, opened 2019-05-14
LLVM 8

### LINKED ISSUE #65: LLVM 8 binding generation fails on macOS
This happens on the Azure build server which currently runs on Mac OS X 10.13 and on Mojave 10.14 when LLVM 8 is installed via `brew install llvm`.

`rust-skia/skia-bindings/skia/include/private/../private/SkOnce.h:11:10: fatal error: 'atomic' file not found`

Not sure why, the Skia build is successful.

pragmatrix:
The Azure Linux and macOS images seem to have LLVM 8 preinstalled and recent builds went fine. So I think we can remove the additional installation steps and recommend LLVM 8 in the readme ([rendered](https://github.com/rust-skia/rust-skia/blob/llvm-8/README.md)).

**Update**: No, Ubuntu builds with 6.0.1 and macOS with Apple LLVM 10. I was confused because of the version output of rustc which prints the LLVM version itself was built with, so I've added the output of `clang --version` as a pipelines step to this PR.

Closes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,13 @@ This project attempts to provide _up to date_ safe bindings that bridge idiomati
 
 Note that the information in this section is preliminary. Please open an issue for any build problem.
 
+### Prerequisites
+
 This project requires LLVM, Python 2, and git to build.
 
-To test if LLVM is installed with the correct version, use `clang --version`. Currently, version 8 is recommended, and at least 7 needs to be available, or - on macOS X - Apple LLVM version 10 should do, too.
+To see which version of LLVM/Clang is available, use `clang --version`. 
+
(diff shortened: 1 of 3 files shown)


pragmatrix:
Perhaps we should just keep the default 6.0.1 for ubuntu 16, Apple LLVM 10 on macOS and install 8.0.0 (the recommended one) for Windows?

For a final decision, I think it's best to block this PR until #116  is resolved to get some feedback about the layout consistency of the generated bindings on different platforms with different LLVM versions.

pragmatrix:
Interestingly, the "brew install llvm" on the macOS image had no effect because the paths were not set, instead, Apple LLVM 10 was used. I've removed that line.

pragmatrix:
Even though we have no data yet on how the different LLVM versions affect the layout of the bindings, I will unblock this because the result should be "better than before". 

Ready for a review.

PR #129, opened 2019-06-22
Add Android support

DenisKolodin:
Hi! I've added minimal support of **Android** platform :tada:
And tested! It works for my `aarch64` with r18b Android NDK. To use it you have to:

1. Download r18b NDK here: https://developer.android.com/ndk/downloads/older_releases.html
2. Create toolchain for compilation: `build/tools/make_standalone_toolchain.py --arch arm64 --install-dir /tmp/ndk`
3. Set `ANDROID_NDK` to r18b folder for `skia` and `bindgen`
4. Compile your code for `aarch64-linux-android` target with:

```
ANDROID_NDK=~/path/to/android-ndk-r18b PATH=$PATH:/tmp/ndk/bin cargo build --target aarch64-linux-android
```

_Notes:_

- It doesn't work for the latest 19 NDK, because Skia doesn't support it yet.
- This PR doesn't cover all cases, but it's the first step toward that.

### DIFF
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -1,6 +1,6 @@
 //! Full build support for the Skia library, SkiaBindings library and bindings.rs file.
 
-use crate::build_support::{cargo, vs};
+use crate::build_support::{android, cargo, vs};
 use bindgen::EnumVariation;
 use cc::Build;
 use std::env;
@@ -144,6 +144,15 @@ impl FinalBuildConfiguration {
                 ("cxx", quote("clang++")),
             ];
 
+            let target = cargo::target();
+            if target.system == "android" {
+                // TODO Not remove, but tweak skia parameters above
+                // for android
+                args.clear();
+                args.push(("cc", quote("clang")));
+                args.push(("cxx", quote("clang++")));
+            }
+
             // further flags that limit the components of Skia debug builds.
             if !build.skia_release {
                 args.push(("skia_enable_atlas_text", no()));
@@ -177,6 +186,13 @@ impl FinalBuildConfiguration {
                     if let Some(win_vc) = vs::resolve_win_vc() {
                         args.push(("win_vc", quote(win_vc.to_str().unwrap())))
                     }
+                }
+                (mut arch, _, "android", _) => {
+                    args.push(("ndk", quote(&android::ndk())));
+                    if arch == "aarch64" {
+                        arch = "arm64";
+                    }
+                    args.push(("target_cpu", quote(arch)));
                 }
                 _ => {}
             }
@@ -252,6 +268,9 @@ impl BinariesConfiguration {
                 link_libraries.extend(vec![
                     "usp10", "ole32", "user32", "gdi32", "fontsub", "opengl32",
                 ]);
+            }
+            (_, _, "android", _) => {
+                link_libraries.extend(vec!["log", "android", "EGL", "GLESv2"]);
             }
             _ => panic!("unsupported target: {:?}", cargo::target()),
         };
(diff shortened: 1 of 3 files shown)


pragmatrix:
THIS _IS_ AWESOME. Thank you @DenisKolodin 

I've tried a build on Windows, Skia builds, but it failed in the binding compilation step (cc_rs) not finding the aarch64 clang even though it was in the path, I'm not sure why, but Windows is always a _bit special_ so this can be postponed for now.

On Ubuntu 18, Skia builds, but I am getting: 
```
/tmp/android-ndk-r18b/sysroot/usr/include/linux/types.h:21:10: fatal error: 'asm/types.h' file not found
```
in the binding generation step. I'll have to take a closer look at that later.

Also, I will try with macOS Mojave later.

One question for now:

Why did you clear all of the gn arguments, causing to build a debug version of _all_ of Skia (>3300) files?

pragmatrix:
I've managed to build skia-bindings on Mojave, but not on High Sierra (also not _that_ important), but on Mojave, linking the tests of skia-safe fails with an ld error `unknown option --as-needed`, probably because the wrong linker is used.

And I've also found out why Skia does not build with the release settings (`is_official_build`, etc): somehow GN tries to use the system's installed freetype, which would not work anyway because of the different architecture I assume. Adding a `args.push(("skia_use_system_freetype2", no()));` is sufficient to build a Skia release version with the limited feature set.

skia-bindings/build_support/skia.rs
```diff
@@ -144,6 +144,15 @@ impl FinalBuildConfiguration {
...
             ];
 
+            let target = cargo::target();
+            if target.system == "android" {
+                // TODO Not remove, but tweak skia parameters above
+                // for android
+                args.clear();
+                args.push(("cc", quote("clang")));
+                args.push(("cxx", quote("clang++")));
+            }
```
pragmatrix:
```rust
if target.system == "android" {
  args.push(("skia_use_system_freetype2", no()));
}
```
This should result in a Skia release build with the limited feature set.

pragmatrix:
To tell rust about the android linker, a `.cargo/config` needs to exist:

```
[target.aarch64-linux-android]
linker = "aarch64-linux-android-clang"
```

Now I am getting 

```
/Users/armin/dev/rust-skia/target/aarch64-linux-android/debug/deps/libskia_bindings-d07a54f14392b5ed.rlib(libskia.SkFontMgr.o): In function `SkFontMgr::RefDefault()::$_0::operator()() const':
          SkFontMgr.cpp:(.text._ZZN9SkFontMgr10RefDefaultEvENK3$_0clEv+0x5c): undefined reference to `SkFontMgr::Factory()'
```
... hinting that we need to tell Skia which fontmanager to compile.

DenisKolodin:
@pragmatrix 
> Why did you clear all of the gn arguments, causing to build a debug version of all of Skia (>3300) files?

I did that temporary, because I couldn't find a combination of parameters to build skia for that target. It worked only with defaults for me. If I activate provided parameters than building fails.

> Adding a args.push(("skia_use_system_freetype2", no())); is sufficient to build a Skia release version with the limited feature set.

Ok, I'll try that  :slightly_smiling_face: :+1:

pragmatrix:
@Brooooooklyn fyi: I am considering merging this and the I'll try to get an Azure job configured to build _and_ link the Android target on macOS. Also I've changed the release process slightly (only the `release` branch will produce binaries and crates, as now documented [in the wiki](https://github.com/rust-skia/rust-skia/wiki#repository-organization)).

In addition to that I'll plan to change the license to append the line "... and rust-skia contributors" to the copyright section and will add handles of people with significant contributions to the readme.

pragmatrix:
License change: 41420f6c4b06d1d1bc244fe69b1c7b33da6e89a2

Brooooooklyn:
@pragmatrix It's great to see other people contributing to our project 😃

PR #141, opened 2019-07-09
Support feature "shaper": Text shaping using HarfBuzz and ICU

pragmatrix:
This PR attempts to wrap Skia's text shaping support.

TODO:

- [x] Compiles on Windows, macOS, and Linux.
- [x] Wrap SkShaper.h
- [x] Add an example.
- [x] Update readme ([rendered](https://github.com/rust-skia/rust-skia/blob/text-shaping/README.md)).
- [x] Bump Versions (14.1).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,4 +5,4 @@ project (skia-bindings)
 
 include_directories(skia-bindings/skia)
 
-add_library(skiabindings skia-bindings/src/bindings.cpp)
+add_library(skiabindings skia-bindings/src/bindings.cpp skia-bindings/src/shaper.cpp)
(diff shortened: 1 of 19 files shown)


pragmatrix:
Here is the small example, I've added to the skia-org console application:

Before (unshaped):

rtl-unshaped

Now with Harfbuzz & ICU: 

rtl-shaped

The example text taken from https://en.wikipedia.org/wiki/Complex_text_layout

And for the first time, this PR contains Skia wrapper code that calls Rust from a C++ class that overrides virtual methods. For that, I am passing a `TraitObject` and a number of Rust `extern "C"` function pointers to C++ and construct an instance that overrides the virtual functions of the base class and calls the Rust functions with the trait object as the first parameter.

pragmatrix:
@Brooooooklyn Care to take a look? May be some ideas to simplify the TraitObject callback mechanism? 

Also - if that's fine with you - I'd like to make this version the first one to be published on crates.io closing #152.

Brooooooklyn:
Happy to see the first publish👍

PR #159, opened 2019-08-05
Crates.io preparations

pragmatrix:
This PR contains some some preparations that are needed for publishing the crates to crates.io.


pragmatrix:
I need to merge this and continue on the master branch for now because the commit hashes of skia and depot-tools are pulled in by cloning the full repository directly from the master branch and merging this PR may change the commit hash that's used in the crate.

PR #165, opened 2019-08-07
[Experiment, DO NOT MERGE] Embed ninja

pragmatrix:
This is an experimental PR to see if it's possible to embed ninja.

I don't actually want to embed executables into the repository, but that PR may offer a way for getting rid of the full depot_tools clone when Skia needs to be built from within the crate.


pragmatrix:
Although we need the ninja executables only, I decided to download the full depot_tools archive for the time being. #167 follows up on that.

PR #185, opened 2019-08-18
Experiment: Support Skia debug builds

pragmatrix:
For Rust Debug builds, Skia was always built in Release mode before (`NDEBUG` defined) which removed all `SkAssert` invariant checks. 

This PR tries to build Skia in Debug mode when `skia-bindings` is built without the `--release` argument.

Even if this works, there will be no changes regarding the prebuilt binaries: when a downloaded cargo package `skia-bindings` is compiled in Debugging mode, the Release binaries will downloaded and linked to for now, so this change is only relevant for developers working with the repository.

### DIFF
--- a/azure-pipelines-build-target.yml
+++ b/azure-pipelines-build-target.yml
@@ -9,8 +9,6 @@ parameters:
   exampleArgs: ''
   # Set up Android environment? Not compatible with runBinaries or runClippy!
   androidEnv: false
-  # Build Skia with Debugging enabled?
-  skiaDebug: '0'
 
 steps:
   - bash: |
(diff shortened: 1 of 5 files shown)


pragmatrix:
After all this time, it finally works to build Skia in debug mode on all platforms. 

To make this available, I am considering to introduce an environment variable, because I fear that Skia built with debugging enabled by default will be detrimental for prototyping Rust applications because of the reduced performance. 

Assuming that most users of rust-skia won't ever need to build Skia in debug mode, an environment varialbe seems to be the sensible choice. If my assumptions about the performance degradation are incorrect, we can change this later on and make debug builds the default for non  `--release` Rust builds.

One thing though. 

I'd like to test the debug builds on the CI for every PR with all features enabled, at least on the primary platforms (no need to build Android and iOS variants I guess). 

And to avoid adding more tasks and time to get the PRs through the CI, I am considering to remove the stable builds - the ones without features - and always build with all features enabled for PRs and the master branch. Build problems in other feature configurations will be catched when the binaries are produced in the release branch, and build problems with no features enabled probably much earlier in local builds.

pragmatrix:
TODOs to finalize this PR:

- [ ] Introduce the environment variable to build Skia in debug mode.
- [ ] Update readme.
- [ ] Azure:
  - [ ] Remove the stable no features builds from the CI
  - [ ] Add an features builds that builds Skia in debug mode.

PR #187, opened 2019-08-20
Pass some reference counted types by value

pragmatrix:
The Skia API already suggested this by declaring a lot of functions such that they expect mutable pointers to reference counted types.

These functions were meant to consume the objects passed in without changing reference counters, and when looking at the examples, this change avoids a large number of borrows and makes the combination of filters feel more natural, for example.

For other types that are reused more frequently, like `Typeface` for example, I extended the function signatures so that they accept a reference or a value.

### DIFF
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -1664,8 +1664,8 @@ extern "C" SkDrawable* C_SkDrawable_Deserialize(const void* data, size_t length)
     return SkDrawable::Deserialize(data, length).release();
 }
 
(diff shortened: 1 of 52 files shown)


pragmatrix:
Additional changes in this PR:
- Introduce and use `RefHandle<N>` for types that are allocated on the heap and freed by Skia.
- `to_option()` was renamed in `into_option()` to communicate that the value converted is consumed.
- All the examples were cleaned up:
  - Pass by value for all filter / shader types.
  - Chained more functions invocations of `Canvas`, `Paint`, and `Path`.
  - replaced `PaintStyle` with `paint::Style`
  - Use `&Path` instead of `&PathBuf` whenever possible.
  -

PR #226, opened 2019-12-04
include! bindings.rs

### LINKED ISSUE #10: Dont generate the bindings in src/bindings.rs

pragmatrix:
Now that IntelliJ Rust has landed support for `include!(concat!(env!("OUT_DIR"), ..))` it's time to prepare a PR that generates the `bindings.rs` file in the output directory and includes it from `skia-bindings/src/lib.rs`.

To enable code insight in IntelliJ based IDEs, the experimental feature `org.rust.cargo.fetch.out.dir` has to be enabled. More information [is available from the IntelliJ Rust team](https://blog.jetbrains.com/clion/2019/12/recent-updates-in-rust/).

I'll keep this PR open and unmerged until both IntelliJ Rust and Rust Analyzer are officially resolving `OUT_DIR` [through the build script output](https://github.com/rust-lang/cargo/pull/7622) as a faster alternative compared to `cargo --build-plan` that is currently in use.

Closes #10

### DIFF
--- a/skia-bindings/.gitignore
+++ b/skia-bindings/.gitignore
@@ -1,2 +1,0 @@
-/src/bindings.rs
-/src/bindings.rs.bk
--- a/skia-bindings/Cargo.toml
+++ b/skia-bindings/Cargo.toml
@@ -21,10 +21,8 @@ include = [
 	"build_support/**/*.rs", 
 	"src/**/*.h", 
 	"src/**/*.cpp", 
-	"src/defaults.rs", 
-	"src/icu.rs", 
-	"src/impls.rs",
-	"src/lib.rs" ]
+	"src/*.rs" 
+]
 
 [lib]
 doctest = false
--- a/skia-bindings/build.rs
+++ b/skia-bindings/build.rs
@@ -1,8 +1,8 @@
 mod build_support;
 use build_support::{binaries, cargo, git, skia, utils};
+use std::io;
 use std::io::Cursor;
 use std::path::Path;
-use std::{fs, io};
 
 /// Environment variables used by this build script.
 mod env {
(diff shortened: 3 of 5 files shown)


pragmatrix:
Here an update regarding the blockers for the PR:

# Rust Analyzer 

has full support, but it must be enabled in the VSCode settings first:

```json
{
    "rust-analyzer.cargoFeatures.loadOutDirsFromCheck": true,
}
```
but Goto Definition is not yet working correctly: 

https://github.com/rust-analyzer/rust-analyzer/issues/3767

# IntelliJ Rust

`org.rust.cargo.fetch.out.dir` must be set in "Experimental Features" (available through the Actions menu).

Otherwise it seems to work fine, so no blockers for IntelliJ Rust anymore.

pragmatrix:
The rust-analyzer team closed https://github.com/rust-lang/rust-analyzer/issues/3767.

pragmatrix:
Closed in favor of #798

PR #227, opened 2019-12-10
Attempt to add all features build

Kethku:
Attempt at fixing: https://github.com/rust-skia/skia-binaries/issues/1

I figure a good middle ground while still being useful would be to publish a build which contains all of the features enabled. Its over kill for my use case, but probably the best bang per buck over building and publishing every sub combination of features.

I'm not sure how to test this without doing an actual release and changing more things. Any guidance?

### DIFF
--- a/azure-pipelines-template.yml
+++ b/azure-pipelines-template.yml
@@ -31,7 +31,10 @@ jobs:
           toolchain: stable
           features: 'textlayout'
           exampleArgs: ''
-
+        stable-all-features:
+          toolchain: stable
+          features: 'vulkan,svg,shaper,textlayout'
+          exampleArgs: ''
       ${{ if eq(parameters.deployRelease, 'False') }}:
         stable-all-features:
           toolchain: stable


pragmatrix:
Good idea, and right in time for the Milestone 79 release. I'll merge this into `master` and `release` to see if the binaries are produced like expected.

As soon that went through, I'll try to deploy the 0.21 release to crates.io so that you can use the prebuilt binaries of skia-safe. You'll probably have to patch that into skulpin's Cargo.toml or create a PR against skulpin, because it's currently not possible to create new binaries for skia-safe crates that were released before.

Kethku:
Awesome! Thanks a bunch for doing this

pragmatrix:
Version 0.21 got published on crates.io and I've tested the binaries download on windows with a small example application. 

Due to the implicit inclusion of the feature `shaper`, the binaries don't include the `shaper` feature id, but both feature configurations `vulkan,svg,shaper,textlayout` and `vulkan,svg,textlayout` should download the correct "all-features" binary.

PR #241, opened 2020-01-01
Add TextStyle::font_size

messense:


pragmatrix:
Nice catch, and thank you!

PR #245, opened 2020-01-05
TypefaceFontProvider Into FontMgr

velyan:
I was trying to load custom fonts for text layout using SkParagraph by the example of [TestFontCollection](https://github.com/google/skia/blob/d445e2b6fcf1a051049b2e527feb1fca9a0a4946/modules/skparagraph/utils/TestFontCollection.cpp#L32) and I couldn't do it with the current API.

```
let mut manager = TypefaceFontProvider::new();
manager.register_typeface(typeface, Some("AlArabiya"));
font_collection.set_asset_font_manager(manager);
```

To get the above working I added an impl for the Into trait on TypefaceFontProvider.

Let me know if there is a better way!

### DIFF
--- a/skia-safe/src/modules/paragraph/typeface_font_provider.rs
+++ b/skia-safe/src/modules/paragraph/typeface_font_provider.rs
@@ -69,6 +69,12 @@ impl Default for RCHandle<sb::skia_textlayout_TypefaceFontProvider> {
     }
 }
 
+impl Into<FontMgr> for TypefaceFontProvider {
+    fn into(self) -> FontMgr {
+        self.deref().clone()
+    }
+}
+
 impl RCHandle<sb::skia_textlayout_TypefaceFontProvider> {
     pub fn new() -> Self {
         Self::from_ptr(unsafe { sb::C_TypefaceFontProvider_new() }).unwrap()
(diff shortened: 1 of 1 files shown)


pragmatrix:
Thank you @velyan 

Although I don't like the "convenience" aspect of this addition, I think we should keep it. 

But for consistency, I suggest that `impl Into<FontMgr> for TypefaceFontProvider` should be added also. And if that would be the only `Into` implementation, the invocation would need to look like `font_collection.set_asset_font_manager(Some(manager.into()))` , which I would prefer, because it makes the conversion and the `Option` explicit.

Also please add at least one testcase for the invocation patterns at the bottom of the same file with a reference to this PR, so that we can try to preserve them in future updates, like so:

```rust
    #[test]
    #[serial_test_derive::serial]
    fn treat_font_provider_as_font_mgr() {
        let mut font_collection = FontCollection::new();
        let typeface = Typeface::default();
        let mut manager = TypefaceFontProvider::new();
        manager.register_typeface(typeface, Some("AlArabiya"));
        font_collection.set_asset_font_manager(manager);
    }
```

velyan:
Thanks @pragmatrix done the change!

Brooooooklyn:
@velyan thanks for your pull request, I think it would be better if you squash your commits

velyan:
@Brooooooklyn squashed, thanks

pragmatrix:
@velyan looks great, I'm merging this.

PR #249, opened 2020-01-09
Enable PDF font subsetting with HarfBuzz

### LINKED ISSUE #248: PDF font subsetting
https://github.com/rust-skia/rust-skia/blob/641b03789b71b66a716df19343dfcf7e0cb825bb/skia-bindings/build_support/skia.rs#L148

I'd like to be able to use sfntly to subset fonts, but it's currently disabled.

messense:
Fixes #248

### DIFF
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -144,6 +144,7 @@ impl TextLayout {
                 ("skia_use_icu", yes()),
                 ("skia_use_system_icu", no()),
                 ("skia_use_harfbuzz", yes()),
+                ("skia_pdf_subset_harfbuzz", yes()),
                 ("skia_use_system_harfbuzz", no()),
                 ("skia_use_sfntly", no()),
             ]);
--- a/skia-bindings/skparagraph.patch
+++ b/skia-bindings/skparagraph.patch
@@ -26,3 +26,28 @@ index 5a85d37c4e..a3ab9b8846 100644
      public_configs = [ ":public_config" ]
      public = skparagraph_public
      if (skia_use_icu && skia_use_harfbuzz) {
+
+diff --git a/modules/skshaper/BUILD.gn b/modules/skshaper/BUILD.gn
+index a672d984f..a8b8bb39d 100644
+--- a/modules/skshaper/BUILD.gn
++++ b/modules/skshaper/BUILD.gn
+@@ -19,6 +19,7 @@ if (skia_enable_skshaper) {
+   }
+
+   component("skshaper") {
++    complete_static_lib = false
+     check_includes = false
+     import("skshaper.gni")
+     public_configs = [ ":public_config" ]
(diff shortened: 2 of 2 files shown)


pragmatrix:
Looks like networking problems on Azure. I am rerunning the tests.

pragmatrix:
While the Windows builds can't download LLVM 9 from chocolately, which is most likely their problem and will be fixed soon, the Linux and macOS builds fail because of linker errors. 

Enabling pdf subsetting embeds the icu library into the skia.lib growing by about 14mb. This in turn causes duplicated symbols, because the skshaper.lib already contains the icu library.

I do have a idea to fix that and will take a closer look later this day.

messense:
@pragmatrix Could you elaborate on how to fix the build issue? I could give it a try.

pragmatrix:
@messense Highly appreciated, I was able to buid Skia with PDF font subsetting with the following patch on MacOS:

```patch
diff --git a/modules/skshaper/BUILD.gn b/modules/skshaper/BUILD.gn
index a672d984f..a8b8bb39d 100644
--- a/modules/skshaper/BUILD.gn
+++ b/modules/skshaper/BUILD.gn
@@ -19,6 +19,7 @@ if (skia_enable_skshaper) {
   }

   component("skshaper") {
+    complete_static_lib = false
     check_includes = false
     import("skshaper.gni")
     public_configs = [ ":public_config" ]
diff --git a/third_party/icu/BUILD.gn b/third_party/icu/BUILD.gn
index 50157f0bf..db72569f6 100644
--- a/third_party/icu/BUILD.gn
+++ b/third_party/icu/BUILD.gn
@@ -68,6 +68,7 @@ if (skia_use_system_icu) {
   }

   third_party("icu") {
+    complete_static_lib = false
     public_include_dirs = [
       "../externals/icu/source/common",
       "../externals/icu/source/i18n",
```

I think this might need to be added to the skshaper.patch in the skia-bindings package to get the build working.

pragmatrix:
Also related is the Skia build flag `is_component_build`.

messense:
It works.

pragmatrix:
Yes, that's very encouraging, I am doing some more builds to see how the size of the binaries change and if the shaper only feature build will work, too.

pragmatrix:
For the record: In `textlayout` feature builds the `icu.lib` is embedded into the `skia.lib`, with `shaper` only feature builds it stays in `skshaper.lib`. Both configurations should be compatible with the prebuilt binaries but must be tested individually before the next release. If anything gets more complicated than that, I am considering to remove the "shaper" feature, because I assume that everyone will use skparagaph / textlayout anyway.

pragmatrix:
`shaper` only builds seem to work, and the size of release build binaries increases about 280kb (tested on MacOS), but only if the textlayout functions are actually used. Given that the resulting executables are usually in the 20mb range, I think that it is reasonable to add PDF subsetting by default instead of adding a new feature flag. 

So I am merging this. Thank you @messense.

PR #262, opened 2020-01-13
Remove bz2 from linking libraries

messense:
I don't think it is used. Tested on linux locally.


pragmatrix:
Thanks! 

The Windows builds on the CI are unrelated, probably a temporary problem with the images or the LLVM at chocolately.

PR #263, opened 2020-01-13
Fix Azure CI Windows builds on image 20200102.1

pragmatrix:


pragmatrix:
For the record: Installing Rust over the preinstalled Rust version of the image 20200102.1 caused the default build toolchain set to `i686-pc-windows-msvc` (32-bit) and subsequently bindgen complained when it found only 64 bit libclang DLLs.

PR #268, opened 2020-01-16
Remove the GL library dependency on Linux

### LINKED ISSUE #267: Is there any way to build rust-skia without libGL dependency?
Or could rust-skia link to a static libGL library instead of system's shared dylib?

pragmatrix:
Since Skia is resolving the OpenGL library and its entry points dynamically, I don't see why this is needed.

Closes #267


pragmatrix:
Wrong, the functions `glXGetCurrentContext` and `glXGetProcAddress` are required by Skia, but they did not need to be linked for the icon and hello example, because release builds do dead code elimination.

Closing this, because the only proper solution seems to support a configuraton that does not use OpenGL at all.

PR #269, opened 2020-01-18
[M80] Support to build Skia with CPU rendering only

### LINKED ISSUE #267: Is there any way to build rust-skia without libGL dependency?

pragmatrix:
This PR changes the default build to a non-GPU build and adds a feature named `gl` that compiles Skia with OpenGL / EGL GPU support (the former default).

This can be interesting for users that need only use Vulkan without OpenGL or render with the CPU only.

Closes #267


messense:
Confirmed that this PR works nicely for my use case.

PR #272, opened 2020-01-20
Pass a FontCollection ref when creating a ParagraphBuilder

velyan:
Does this change make sense or am I missing something? 
Is it okay to share a FontCollection reference across ParagraphBuilders?

### DIFF
--- a/skia-safe/examples/skia-org/skparagraph_example.rs
+++ b/skia-safe/examples/skia-org/skparagraph_example.rs
@@ -24,7 +24,7 @@ fn draw_lorem_ipsum(canvas: &mut Canvas) {
         font_collection.disable_font_fallback();
 
         let paragraph_style = ParagraphStyle::new();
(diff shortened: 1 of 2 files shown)


pragmatrix:
I must admit that it looks wrong, because from the signature of the function it indicates that the ownership of `FontCollection` is passed to Skia. 

But with the change in this PR, the reference counter would stay at 1 when `FontCollection` is passed, even though there are now two references existing, the one in the builder and the one on the Rust side.

To clarify further, a `FontCollection` is a `RCHandle<>`, a cheap-to-clone Skia shared pointer reference counted object, for which cloning does not mean duplication but that a mutable shared Rust handle is returned that points to the same object. What this means is that - even though a bit unsafe - rust-skia supports single threaded shared mutability, which is comparable to a `Rc<RefCell<T>>` type in Rust.

TLDR. To reuse and share `FontCollection`, it must be pass to the builder with `.clone()`.

velyan:
Oh I see, thanks for explaining this @pragmatrix!

PR #275, opened 2020-01-23
Support Metal

pragmatrix:
This PR is an attempt to support Metal for macOS and iOS targets.


pragmatrix:
A Proof of Concept seems to work. 

I won't be updating the CI for building metal in this PR, because I actually have no idea how we can configure it to build the proper feature combinations yet.

- [x] Builds macOS target with feature `metal` with Xcode / Apple LLVM 11.
- [x] Builds iOS target with feature `metal` with Xcode / Apple LLVM 11.
- [ ] Test with LLVM 9 on Catalina
  - [ ] macOS
  - [ ] iOS
- [ ] Create and submit another PR that updates build support on Catalina.
- [ ] Add more bindings to get parity with vulkan and opengl.

Issue #290, opened 2020-02-16
Thread-safety and Send's


ostera:
Hi!

I've started to write some bindings for Erlang to Skia through this library, and I quickly hit a blocker with structs not implementing Send.

I'm not a Skia expert, but [from](https://groups.google.com/forum/#!searchin/skia-discuss/thread$20safe%7Csort:date/skia-discuss/TUy0I596FlM/dB8ul_s3CAAJ) [reading](https://groups.google.com/forum/#!topic/skia-discuss/-G1cyl1QD9E) [around](https://www.google.com/search?q=site%3Aapi.skia.org+thread+safe) I've found that many of them (with exceptions like at least `SkTypeface` and `SkBitmap`) should be safe to implement Send on under most circumstances.

Do you think it would be a good idea to do this sort of work? I'd be happy to volunteer some time onto it.

pragmatrix:
Hello @ostera!

> Do you think it would be a good idea to do this sort of work?

Yes, of course!

Some thoughts:

- I think we can assume that all reference counted types are safe to `Send` when their reference count is 1 and do not refer to other (sharable) types. We could solve that in a similar manner like [`Rc::try_unwrap`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.try_unwrap) returning a wrapper that is `Send` and can be converted to `RCHandle<T>` in another thread. Depending on your use-case that may already solve what you need.
- An `RCHandle<T>` that is directly `Send`, must be also be `Sync`. Because the handle could be cloned and then sent to another thread pointing to the same object, effectively resulting in a handle that is `Sync`.
- All immutable and thread safe types like `SkTypeface` can be `Send + Sync` directly on their wrapper type.
- For some types, there is documentation available in the Skia sources, which we can follow up on, for example:
  - `SkTypeface`:
    > Typeface objects are immutable, and so they can be shared between threads.
  - `SkBitmap`:
    > SkBitmap is not thread safe. Each thread must have its own copy of SkBitmap fields,
    although threads may share the underlying pixel array.
- Sometimes it's interesting to take a look at the implementation. For example, `SkTypeface` is not actually immutable, but maintains the illusion by using thread synchronization to compute values lazily.

ostera:
@pragmatrix, fantastic! :tada:

[...]

PR #291, opened 2020-02-16
Improve Thread-safety

ostera:
Hi! This PR refers to #290. I've started it in a rather lenient way, making _everything_ a `Send+Sync` by running two sed commands.

The final list of things that _should_ and **should not** be thread-safe is likely different, and so is the way in which they should be. I'm mostly hacking this to unblock the Erlang bindings I'm experimenting with.

Here's the list of types handles that were extended and we should revisit:

<details>

| DataType | Is thread-safe | Fixed in PR |
|--------|--------------------|------------------|
| BackendDrawableInfo |  Yes  |  Yes  |
| BackendFormat |  Yes  |  Yes  |
| BackendRenderTarget |  Yes  |  Yes  |
| BackendTexture |  Yes  |  Yes  |
| BBHFactory |  Yes  |  Yes  |
| BiDiRunIterator |  Yes  |  Yes  |
| Bitmap |  Yes  |  Yes  |
| Builder |  Yes  |  Yes  |
| ColorFilter |  Yes  |  Yes  |
| ColorInfo |  Yes  |  Yes  |
| ColorMatrix |  Yes  |  Yes  |
| ColorSpace |  Yes  |  Yes  |
| Context |  Yes  |  Yes  |
| ContourMeasureIter |  Yes  |  Yes  |
| ContourMeasure |  Yes  |  Yes  |
| Data |  Yes  |  Yes  |
| DataTable |  Yes  |  Yes  |
| DeferredDisplayListRecorder |  Yes  |  Yes  |
| DeferredDisplayList |  Yes  |  Yes  |
| Drawable |  Yes  |  Yes  |
| DrawLooper |  Yes  |  Yes  |
| DynamicMemoryWStream |  Yes  |  Yes  |
| Extensions |  Yes  |  Yes  |
| FontCollection |  Yes  |  Yes  |
| Font |  Yes  |  Yes  |
| FontMgr |  Yes  |  Yes  |
| FontRunIterator |  Yes  |  Yes  |
| FontStyleSet |  Yes  |  Yes  |
| GPUDrawHandler |  Yes  |  Yes  |
| ImageFilter |  Yes  |  Yes  |
| ImageGenerator |  Yes  |  Yes  |
| ImageInfo |  Yes  |  Yes  |
| Image |  Yes  |  Yes  |
| Interface |  Yes  |  Yes  |
| InternalMetadata |  Yes  |  Yes  |
| Interpolator |  Yes  |  Yes  |
| LanguageRunIterator |  Yes  |  Yes  |
| LineMetricsVector<'a> |  Yes  |  Yes  |
| LocalizedStringsIter |  Yes  |  Yes  |
| MaskFilter |  Yes  |  Yes  |
| OpBuilder |  Yes  |  Yes  |
| Paint |  Yes  |  Yes  |
| ParagraphBuilder |  Yes  |  Yes  |
| ParagraphCache |  Yes  |  Yes  |
| Paragraph |  Yes  |  Yes  |
| ParagraphStyle |  Yes  |  Yes  |
| PathEffect |  Yes  |  Yes  |
| Path |  Yes  |  Yes  |
| PathMeasure |  Yes  |  Yes  |
| Picture |  Yes  |  Yes  |
| PictureRecorder |  Yes  |  Yes  |
| PixelRef |  Yes  |  Yes  |
| Pixmap |  Yes  |  Yes  |
| Region |  Yes  |  Yes  |
| ScriptRunIterator |  Yes  |  Yes  |
| Shader |  Yes  |  Yes  |
| Shaper |  Yes  |  Yes  |
| String |  Yes  |  Yes  |
| Strings |  Yes  |  Yes  |
| StrokeRec |  Yes  |  Yes  |
| StrutStyle |  Yes  |  Yes  |
| SurfaceCharacterization |  Yes  |  Yes  |
| Surface |  Yes  |  Yes  |
| TextBlobBuilder |  Yes  |  Yes  |
| TextBlobIter<'a> |  Yes  |  Yes  |
| TextBlob |  Yes  |  Yes  |
| TextBoxes |  Yes  |  Yes  |
| TextStyle |  Yes  |  Yes  |
| TextureInfo |  Yes  |  Yes  |
| TypefaceFontProvider |  Yes  |  Yes  |
| TypefaceFontStyleSet |  Yes  |  Yes  |
| Typeface |  Yes  |  Yes  |
| Typefaces |  Yes  |  Yes  |
| Vertices |  Yes  |  Yes  |
| View3D |  Yes  |  Yes  |

</details>


ostera:
And so it begins!

first-drawing

pragmatrix:
Nice to see progress,

but I'm wondering why you need Sync at all. An Erlang actor runs on isolated local state and does not support multithreading, so do you need access Skia types from two different actors at the same time? And when does the sharing occur, or in other words, at which point is Sync rather than Send needed?

ostera:
Good question!

I originally only had the Send's in place, but rereading your comment from #290, I figured I might as well just add Sync's too. We can narrow down the ones that are actually sync-able now, removing the trait from them.

Regarding Erlang's concurrency model, any Erlang process will run on any one of the scheduling threads at a time, and we should be able to move data across threads for this to be safe. Send is enough for this too.

It is however possible to share a _resource reference_, the opaque immutable value that points to the scheduler managed memory space that in this case contains a Skia value, with another process. There is no concept of ownership of this data, and if you can get your hands on that ref then it's possible that two processes will try to access the same data at the same time. I need to look deeper into this, since the BEAM may be already doing some read/write locks on refs.

ostera:
starting to be a little more high-level/usable :) 

hello_joe

pragmatrix:
Hello @ostera, I am closing this PR in favor of #362. 

I was going through all types and decided one by one if Send and Sync should be implemented. If you consider to continue on the BEAM port I would really like your feedback, may be I missed something.

ostera:
Absolutely! Thanks for the heads up 🙌🏼

PR #298, opened 2020-03-05
skia-org: point offscreen-rendering-context ref to a repository under…

pragmatrix:
… rust-skia

The current master does not build, because the rev we are pointing to vanished from the original repository. So, until we port skia-org to a replacement for offscreen-rendering-context, I've decided to fork the repository under the rust-skia organization to fix the build in master.


pragmatrix:
This problem does not affect the released crates.

PR #317, opened 2020-04-16
add gl window example

nornagon:
This adds a new example showing how to render directly into an OS window. Uses the GL backend for now because I couldn't get metal working and I don't currently have a setup to test linux or windows for vulkan / dx backends.


pragmatrix:
Thank you a lot, OpenGL only is fine. I am merging this as it is and will try to fix the CI error (which is unrelated) and add some minor tweaks later.

PR #332, opened 2020-05-02
Add binding dl env variable

j4qfrost:


pragmatrix:
Hi, thank you for the PR, I see that you are struggling with the CI. Before you continue, I have two suggestions:

- I think the environment variable should be named `SKIA_BINARIES_URL` 
- I would like to provide a full template that defaults to `https://github.com/rust-skia/skia-binaries/releases/download/{tag}/skia-binaries-{key}.tar.gz` with the named parameters replaced with `String::replace`, because adding another dependency is probably not worth it.

j4qfrost:
Updated, I hope I understood correctly.

pragmatrix:
yes you did, thank you!

skia-bindings/build_support/binaries.rs
```diff
@@ -120,12 +121,11 @@ pub fn key(repository_short_hash: &str, features: &[impl AsRef<str>], skia_debug
 
 /// Create the download URL for the prebuilt binaries archive.
 pub fn download_url(tag: impl AsRef<str>, key: impl AsRef<str>) -> String {
-    format!(
-        "https://github.com/rust-skia/skia-binaries/releases/download/{}/{}-{}.tar.gz",
-        tag.as_ref(),
-        ARCHIVE_NAME,
-        key.as_ref()
-    )
+    let binding_url = env::var("SKIA_BINARIES_URL")
```
pragmatrix, line 124: One more thing, use `cargo::env_var()` here, so that the build script is called again when the environment variable changes.

j4qfrost: Done.

PR #341, opened 2020-06-01
bug fix: borrow pixmap of an image if it was able to peek pixels nati…

lykhonis:
Should borrow pixmap it was able to peek it natively


pragmatrix:
Great, looks reasonable, nice catch! Do you need a crate release with updated binaries?

/azp run

lykhonis:
No need for binaries :) thanks.

PR #342, opened 2020-06-02
- Add support for perform deferred clean up on GrContext

lykhonis:
Expose perform deferred clean up

### DIFF
--- a/skia-safe/src/gpu/context.rs
+++ b/skia-safe/src/gpu/context.rs
@@ -7,6 +7,7 @@ use crate::prelude::*;
 use crate::{image, ColorType, Data, Image};
 use skia_bindings as sb;
 use skia_bindings::{GrContext, SkRefCntBase};
+use std::time::Duration;
 
 pub type Context = RCHandle<GrContext>;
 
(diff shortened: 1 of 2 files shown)


skia-safe/src/gpu/context.rs
```diff
@@ -142,7 +142,15 @@ impl RCHandle<GrContext> {
         self
     }
 
-    // TODO: performDeferredCleanup()
+    pub fn perform_deferred_cleanup(&mut self, not_used_millis: i32) -> &mut Self {
```
pragmatrix, line 145: So far I've preferred `std::time::Duration` to represent untyped time duration values on the Rust side. What do you think, can we ignore the performance a bit to get more type safety here?

lykhonis: Totally. I was going to do that, but then thought just reflect 1:1 the Skia's APIs.

pragmatrix:
The errors on the CI seem to be related to new updated clippy lints in the beta rust compiler that was probably released today. I will fix these errors in another PR.

Issue #340, opened 2020-05-26
Packaging binaries that include skia-bindings is proving difficult.


theotherjimmy:
I'm packaging neovide, https://github.com/Kethku/neovide, that depends on skulpin, and therefore skia-safe and skia-bindings, for NixOS/nixpkgs. Nix, and many other package managers, require that the code fetch phase is separate from the build phases, which do not have a network connection. You can see my progress here: https://gist.github.com/theotherjimmy/d830cc971c9a9c4f0385eb7bd687d814.

This separation has proven quite difficult to handle with skia-bindings, as part of the `bulid.rs` script either pulls binaries from github or tries to sync the skia repos with git, both off which fail. I have been hacking through the `build.rs` script, and have not yet come up with a method for separating these 2 phases. Ideally, I would like to enhance the build scripts here to allow building of rust-skia based binaries without a downstream fork or a patch, as those would likely be difficult to maintain and brittle.

I'm happy to write Rust and submit pull requests to help in this effort, and I would like advice on how to proceed to make it easier to package binaries such as neovide.

pragmatrix:
I think it should be possible to provide another environment variable that points to a fully configured Skia source tree Ninja can build. Of course given that the configuration matches the one the build script computes based on the current crate feature combination and the `SKIA_DEBUG` environment variable. 

The build could compare the `gn` arguments with the one the Rust `build.rs` script would use and issue a warning if they don't match. `gn` writes them into a file at the target root named `args.gn`.

Assuming that you package based on the crate releases, `depot_tools` is also pulled in as a submodule, but only to download the ninja executable itself. I'd suggested to use one single environment variable that we call something like `SKIA_OFFLINE_GN_ROOT` (a better name could probably be found), which also implies that ninja is preinstalled.

theotherjimmy:
Thanks for the advice, and getting back to me so quickly. I think `SKIA_OFFLINE_GN_ROOT` implies that `build.rs` should not do any network-related actions, and should cover the "we already have ninja" case. I think it would be possible to package neovide and other binaries with this env var available.

pragmatrix:
This is getting a bit more complicated that anticipated. 

First, I think we need another environment variable for specifying where the command should invoked from. And then the build script in it's current form always assumes that the GC build root is a subdirectory `skia` in the current target directory of the build. 

PR #343 is a first step in that direction but does not implement the separation of the directories yet.

theotherjimmy:
[...]

PR #343, opened 2020-06-03
Support offline builds

pragmatrix:
An initial rough idea on how offline builds can be implemented like suggested in #340.


theotherjimmy:
Finally looking at this, and it strikes me as odd that we allow a packager to provide their own ninja executable, but we assume that the gn executable will be provided as part off the source. Would it make sense to expect the packager to provide gn as well as ninja?

pragmatrix:
That distinction comes from the fact that ninja was pulled from the depot_tools repository and gn is already contained in Skia's repository, which could lead to the assumption that gn is more tightly coupled to the Skia sources than ninja. But since I don't know how they differ in terms of evolution in relation to the source, the distinction could also be solely justified by that fact that adding another dependency is not needed, because gn must available anyway and is guaranteed to be the correct version to use. 

If you think there is a use case for a separate gn, we should add another environment variable. Although I personally would delay that until an actual use case comes up to get the chance to learn from that.

pragmatrix:
Oh, looking closer at the source I just recognized that git-sync-deps is most probably responsible to fetch the gn executable, which looks like a platform dependent executable (I assumed that it's just a python script). So I think now that you are right, and adding an environment variable is probably the best way to externalize all platform specific dependencies.

pragmatrix:
I've added `SKIA_OFFLINE_GN_COMMAND` environment variable. If set, it should contain the full path to the gn executable that shall be used instead of the default path.

pragmatrix:
I've added some lines of documentation for the environment variables ([rendered](https://github.com/pragmatrix/rust-skia/blob/offline-builds/skia-bindings/README.md#offline-builds)) and would like to merge this as soon as possible to get this MR into the Milestone 84 release.

pragmatrix:
@theotherjimmy Thank you for helping and making offline build support possible. I am merging this now and will close #340 after that, please open new issues if new information comes up.

PR #346, opened 2020-06-06
Attempt to support pc-windows-gnu targets on Windows

### LINKED ISSUE #345: x86_64-pc-windows-gnu support
Right now skia-safe supports only `x86_64-pc-windows-msvc` when building for Windows. This makes it impossible to cross-compile Windows binaries from Linux that depend on skia-safe. Ideally it should also support `x86_64-pc-windows-gnu`, which would make it possible to cross-compile things for Windows that depend on rust-skia.

pragmatrix:
This PR tries to implement support for pc-windows-gnu targets on Windows.

Closes #345

### DIFF
--- a/azure-pipelines-template.yml
+++ b/azure-pipelines-template.yml
@@ -145,6 +145,12 @@ jobs:
       ${{ if and(eq(parameters.platform, 'macOS'), eq(parameters.deployRelease, 'False')) }}:
         features: '$(features),metal'
 
+  - ${{ if eq(parameters.platform, 'Windows') }}:
+    - template: 'azure-pipelines-build-target.yml'
+      parameters:
+        target: 'x86_64-pc-windows-gnu'
+        releaseBinaries: ${{ parameters.deployRelease }}
+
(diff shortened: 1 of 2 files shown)


pragmatrix:
Verdict (by 5.6 Sol), closing.

Supporting x86_64-pc-windows-gnu is feasible, but not as a small rust-skia configuration change. Native static MinGW support requires a maintained Skia fork change.

Current blockers:

Skia hardcodes its MSVC toolchain whenever target_os == "win" in BUILDCONFIG.gn:251.
Windows build flags and SDK paths are MSVC-specific in BUILD.gn:62.
rust-skia routes GNU Windows through windows::Generic in platform.rs:97, but that path cannot override Skia’s toolchain selection.
Archive naming assumes every Windows library is name.lib; MinGW produces libname.a in cargo.rs:83.
The C++ wrapper would be compiled using the GNU target while Skia remains clang-cl/MSVC. Since the wrapper calls Skia C++ symbols, their C++ ABIs are incompatible.

PR #353, opened 2020-07-16
Don't link skia-bindings two times in a full build

pragmatrix:
Although not causing an error so far, skia-bindings was added to list of linker libraries twice if a full build was invoked, once by the invocation of cc and once by our build_support. This PR disables the cargo metadata submission caused by cc.


pragmatrix:
The CI errors are probably caused by a regression in the current cargo version 1.46 beta. I'll take care of that later.

PR #364, opened 2020-08-02
add Data::new_bytes to create without copy

syrel:
It would be nice to be able to create Data without copy. Should `new_bytes` be marked `unsafe`?

### DIFF
--- a/skia-safe/src/core/data.rs
+++ b/skia-safe/src/core/data.rs
@@ -70,9 +70,8 @@ impl RCHandle<SkData> {
     /// Constructs Data from a given byte slice without copying it.
     ///
     /// Users must make sure that the underlying slice will outlive the lifetime of the Data.
-    pub fn new_bytes(data: &[u8]) -> Self {
-        Data::from_ptr(unsafe { sb::C_SkData_MakeWithoutCopy(data.as_ptr() as _, data.len()) })
-            .unwrap()
+    pub unsafe fn new_bytes(data: &[u8]) -> Self {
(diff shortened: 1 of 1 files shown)


pragmatrix:
I am a bit resistant supporting this, for once, it's unsafe. The function itself can be made safe by using `Borrows`, like [here](https://github.com/rust-skia/rust-skia/blob/2deea04e442e46ca48cab7b5d07f7c7bd117b098/skia-safe/src/core/pixmap.rs#L32-L49), but since an `RCHandle<SkData>` can always be cloned out of the `Borrowed` wrapper, it's also always possible to just clone it and dispose of the `Borrowed<>`.

Originally, I did not even intend to wrap SkData, because Rust supports similar constructs through slices and Vecs out of the box. For that reason, I suggest to take first a look on the use cases you'd like to support and if we might find a better way to prevent the unnecessary copies.

pragmatrix:
Thinking about this a bit longer I can't see a way to make this safe, so add the unsafe and we'll just merge this. The comment on the original Skia function is interesting though:

```c++
    /**
     *  Call this when the data parameter is already const and will outlive the lifetime of the
     *  SkData. Suitable for with const globals.
     */
    static sk_sp<SkData> MakeWithoutCopy(const void* data, size_t length) {
        return MakeWithProc(data, length, DummyReleaseProc, nullptr);
    }
```

pragmatrix:
@syrel I've marked the function as unsafe and will merge this PR as soon it gets through the CI.

syrel:
I absolutely agree that it does not perfectly fit skia-**safe** as there indeed no nice way to implement a safe wrapper for this api, hence adding unsafe is a reasonable solution as long as it is documented 👍
Thank you for merging this

Issue #398, opened 2020-08-31
canvas not taken as reference


Kethku:
https://github.com/rust-skia/rust-skia/blob/f4709ab7336a25b3e411c8c3e2ea4331eb3ff2a8/skia-safe/src/core/surface.rs#L380

This line consumes the canvas passed to it. I think that may be incorrect?

Kethku:
I'm an idiot. This is my problem

Kethku:
Actually yes. I believe this is a problem. The canvas should be &mut not just mut.

pragmatrix:
The function should accept both, a canvas by taking ownership, and canvas - the most likely case - by a mut reference. At least that was the original intention, which - I'll admit - might be confusing. 

``` rust
#[test]
fn test_drawing_as_mut_ergonomics() {
    let mut surface = Surface::new_raster_n32_premul((16, 16)).unwrap();

    // option1:
    // - An _owned_ canvas can be drawn by transferring ownership.
    {
        let canvas = Canvas::new(ISize::new(16, 16), None).unwrap();
        surface.draw(canvas, (10.0, 10.0), None);
    }

    // option2:
    // - An &mut canvas can be drawn.

    {
        let mut canvas = Canvas::new(ISize::new(16, 16), None).unwrap();
        surface.draw(&mut canvas, (10.0, 10.0), None);
    }

    // option3:
    // - A canvas from another surface can be drawn.
    {
        let mut surface2 = Surface::new_raster_n32_premul((16, 16)).unwrap();
        let canvas = surface2.canvas();
        surface.draw(canvas, (10.0, 10.0), None);
    }
}
```

See also: https://github.com/rust-skia/rust-skia/blob/94cd93e9e1aa8f6ddca8ae724809a286521c8a6b/skia-safe/src/core/canvas.rs#L180

And I am absolutely not sure if this is in any way idiomatic Rust.

Kethku:
https://github.com/rust-skia/rust-skia/pull/399/files#r481846944 this gives an example closer to my problem. I'm very much not a Rust expert though so there is very possibly a gap in my understanding. Thanks so much for humoring me thus far.

Kethku:
I see your second point about AsMut. So I guess the fix for my issue is to just call as_mut manually rather than passing the mutable reference?

pragmatrix:
[...]

PR #399, opened 2020-08-31
Add API tests to draw canvases on surfaces with AsMut

pragmatrix:
See #398

### DIFF
--- a/skia-safe/src/core/surface.rs
+++ b/skia-safe/src/core/surface.rs
@@ -374,7 +374,6 @@ impl RCHandle<SkSurface> {
         })
     }
 
-    // TODO: why is self mutable here?
     pub fn draw(
         &mut self,
         mut canvas: impl AsMut<Canvas>,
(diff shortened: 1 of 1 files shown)


skia-safe/src/core/surface.rs
```diff
@@ -519,3 +518,31 @@ fn test_raster_direct() {
...
+
+    {
+        let mut canvas = Canvas::new(ISize::new(16, 16), None).unwrap();
+        surface.draw(&mut canvas, (10.0, 10.0), None);
+    }
+
+    // option3:
+    // - A canvas from another surface can be drawn.
+    {
+        let mut surface2 = Surface::new_raster_n32_premul((16, 16)).unwrap();
```
Kethku, line 544: The version I'm struggling with (I think) is like this:

``` Rust
fn test_borrowed(canvas: &mut Canvas) {
    let mut surface = Surface::new_raster_n32_premul((16, 16)).unwrap();
    surface.draw(canvas, (5.0, 5.0), None);
    surface.draw(canvas, (10.0, 10.0), None);
}
```

I'm still learning Rust but my conception is that this function should borrow the canvas mutably, and then release the mutable borrow at the end of the function call. Then in the next draw it will mutably borrow the canvas again. Unfortunately instead I get an error: "move occurs because `canvas` has type `&mut skulpin::skia_safe::Canvas`, which does not implement the `Copy` trait"

pragmatrix: Still learning, too. 

You are right, this undermines what I had intended originally. It seems that the conversion to the `impl AsMut<>` consumes the reference.

PR #402, opened 2020-09-02
Don't use impl AsMut anymore in functions that expect mutable Canvas.

pragmatrix:
Given that it is very confusing (#398) and it actually does not even make sense to take ownership in the case of an `OwnedCanvas`, this PR removes all `impl AsMut<>` parameter signatures.

This change is breaking backwards compatibility, but hopefully only so for some rarely used edge cases.

I am keeping and deprecating the `AsMut<>` trait implementation for `&mut Canvas` as `OwnedCanvas`, so that uses of `.as_mut()` will compile after updating to the version that will contain this change.

@Kethku Would you mind to look at the changes and see if they resonate with what you would expect from the APIs that are  related to the Canvas type? Thanks!

### DIFF
--- a/skia-safe/Cargo.toml
+++ b/skia-safe/Cargo.toml
@@ -11,7 +11,7 @@ keywords = ["skia", "rust-bindings", "vulkan", "opengl", "pdf"]
 categories = ["api-bindings", "graphics", "multimedia::images", "rendering::graphics-api", "visualization"] 
 license = "MIT"
 
-version = "0.33.1"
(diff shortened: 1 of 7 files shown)


Kethku:
these make good sense to me but I'm really not an expert :P

ctrlcctrlv:
I also thought this was strange and am glad you're changing it.

pragmatrix:
Thank you both for the feedback.

PR #426, opened 2020-10-08
replace curl with ureq:

j4qfrost:
Related to this, https://github.com/Kethku/neovide/issues/371

I figured since curl was just being used in one place, we could just use something rust native to relieve the dependency.

### DIFF
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -1123,8 +1123,8 @@ pub(crate) mod rewrite {
     }
 
     pub fn k_xxx(name: &str, variant: &str) -> String {
-        if variant.starts_with('k') {
-            variant[1..].into()
+        if let Some(stripped) = variant.strip_prefix('k') {
+            stripped.into()
         } else {
             panic!(
                 "Variant name '{}' of enum type '{}' is expected to start with a 'k'",
@@ -1356,8 +1356,8 @@ pub(crate) mod definitions {
             defines
                 .split_whitespace()
                 .map(|d| {
-                    if d.starts_with(prefix) {
-                        &d[prefix.len()..]
+                    if let Some(stripped) = d.strip_prefix(prefix) {
+                        stripped
                     } else {
                         panic!("missing '-D' prefix from a definition")
                     }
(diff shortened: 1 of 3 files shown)


skia-bindings/build_support/skia.rs
```diff
@@ -1123,8 +1123,8 @@ pub(crate) mod rewrite {
     }
 
     pub fn k_xxx(name: &str, variant: &str) -> String {
-        if variant.starts_with('k') {
-            variant[1..].into()
+        if let Some(stripped) = variant.strip_prefix('k') {
```
j4qfrost, line 1126: These were some clippy changes. If you would prefer, I can make a separate PR

pragmatrix:
Thanks @j4qfrost, [we decided last year](https://github.com/rust-skia/rust-skia/pull/157) to use curl instead of reqwest to reduce build times and the number of dependencies, but seeing that neovide users may have less problems compiling with native Rust depedencies, we probably should prefer a Rust implementation. 

But compared to master, and without any features, this PR increases the number of dependencies from 97 to 132 (36%), and slow down build times by roughly 50% (59s, 57s vs 1.6m.59 1.5m.34), so I am not so sure if this is the right way to move forward. Perhaps it would be possible to enable ureq behind a feature?

What I also recognized is that in the related neovide issue no reason was provided about why openssl could not be installed.

j4qfrost:
Thanks for the response. Is there a place where I can take a look at the metric? I want to see if I can grind down the compile times for this. If it's not possible I'll put it behind a feature.
The only edge case I can think of with regard to not installing openssl is compilation on a minimal sized docker container, but even then you would just use a builder.

I found something that looks promising https://github.com/jayjamesjay/http_req. It builds in 11.83s on my laptop. I'll swap it out tomorrow.

pragmatrix:
@j4qfrost I think I've misinterpreted the measurements, it seems that the slowdown is only about 10% seconds on my Windows machine. So this would be tolerable for the full build, but would probably get worse when the binaries are being downloaded from the crates.

pragmatrix:
Merging ureq support to see how the adoption together with the upcoming M87 Release #423 works out. Thank you @j4qfrost.

PR #432, opened 2020-12-07
Allow cross-compilation for arbitrary platforms

Vurich:
This PR allows the build script to support cross-compilation for any platform, not just the limited set explicitly matched on. Obviously this isn't anywhere near perfect and there will be plenty of platforms for which this doesn't work, but at the bare minimum it supports compiling for the Raspberry Pi (and other ARM64 CPUs), which I think is a relatively-common platform that people might want to compile for. This PR also includes a really minor commit that I made because it makes it easier to test individual projects inside rust-skia, by excluding _all_ target dirs from git. If you don't want to include that commit or you'd want me to submit a separate pull request I'd be more than happy to oblige.


pragmatrix:
I really like this PR.

> it supports compiling on the Raspberry Pi

Nice!

> excluding all target dirs from git

This is fine.

If the formatting is fixed, we'll see if the CI builds the platforms we already do support.

Some initial observations:

- I think the environment variable `SYSROOT` should be renamed to `SKIA_SYSROOT`. I am not sure about `OPT_LEVEL`, because that seems to be a commonly used convention.
- For retrieving the env vars, the function `cargo::env_var()` in `cargo.rs` should be used (which automatically outputs the `rerun_if_env_changed`).
- I think we should not pass all flags to `extra_asmflags` and `extra_ldflags`, only the ones that are actually needed.

And thank you for updating bindgen.

Vurich:
`OPT_LEVEL` is actually set by Cargo itself, it's not a custom environment variable (hence why there's no `rerun_if_env_changed` for it). You make a good point about `SYSROOT`, but again it's a relatively common variable and there's no reason to set a different sysroot for Skia compared to other libraries. I'm no zealot over it though, I'd be happy to change it to `SKIA_SYSROOT` if you feel strongly about it.

Fair point about `extra_asmflags`/`extra_ldflags`. They actually weren't needed to get it compiling, but the flags were mentioned in the Skia cross-compilation docs so I added them

Vurich:
So the only argument to `extra_asmflags` I need is the target, so I've fixed it so that that's the only argument it passes. I just had the though: do you think the sysroot should be managed as part of this `BuildConfiguration` struct that you have? I don't fully understand how your build utilities work but it might be cleaner to create the `BuildConfiguration` from the `SYSROOT`/`OPT_LEVEL` environment variables and pass them in that way.

pragmatrix:
> You make a good point about SYSROOT, but again it's a relatively common variable

Then we support both environment variables as you suggested.

Regarding `BuildConfiguration`:  It should be extended if a build variable effects the _quality_ and _feature set_ of the build result, so `OPT_LEVEL` probably yes, but `SYSROOT` perhaps not. `on_windows` slipped in perhaps because it was convenient at the time.

Vurich:
Sure, I'll fix it up so `OPT_LEVEL` is included in the build configuration. You can probably see that I've made further changes to the build script, I intended to include these as part of a new PR but GitHub doesn't allow you to change the branch that a PR points to. If you don't want these extra changes I'll create a new PR, but if you're fine with them then I might as well keep using the same PR.

pragmatrix:
I'd prefer separate PRs. It's already hard to see what changes caused the errors on the CI builds.

Vurich:
Fine by me. Especially since one of the changes (forcing users to manually select `x11` as a feature, and otherwise not linking with `libGL`) is breaking.

PR #433, opened 2020-12-09
Allow cross-compilation for arbitrary targets

Vurich:
This allows manually setting the sysroot (instead of only automatically detecting sysroot, and only on macOS). Additionally, it correctly sets the target flag for the compiler on all platforms. There seems to be an env var `SDKROOT` that is used by clang to set sysroot without CLI flags, so I've reused that variable name for consistency.

Duplicated from #432 but moved to a new branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 /Cargo.lock
-/target/
+target/
 /.cargo/credentials
 /cmake-build-debug/
--- a/skia-bindings/Cargo.toml
+++ b/skia-bindings/Cargo.toml
@@ -53,7 +53,7 @@ shaper = ["textlayout"]
 
 [build-dependencies]
 cc = "1.0.37"
-bindgen = "0.55.1"
+bindgen = "0.56.0"
 # For enum variant name replacements.
 regex = "1.2.1"
 heck = "0.3.1"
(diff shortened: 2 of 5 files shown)


pragmatrix:
@Vurich Current master builds on the CI, please rebase this PR to continue.

Vurich:
Still failing due to what appears to be bugs unrelated to my changes. Specifically, linker errors.

pragmatrix:
The linker errors appear only with `--release` builds in this branch but not in master.

Vurich:
My best guess is that it's because I enable LTO and CI has `CC=gcc CXX=g++`, which means that the LTO-enabled artifacts are incorrect. I don't want to disable setting the CC/CXX entirely because it's useful for cross-compilation and for using ccache, but maybe it's better to just keep LTO disabled for now. Maybe eventually there could be heuristic detection of whether the compiler is clang-like or gcc-like (which is how bindgen handles flags), but for now it's fine to just disable LTO.

pragmatrix:
I can confirm the LTO caused the linker errors. I've created a new branch cross-compile-fix-ci where with LTO disabled and no -O on Windows to prevent some warnings. Let's see how it does.

PR #435, opened 2020-12-11
Enable compilation using EGL/Wayland

Vurich:
This allows skia-safe to compile using EGL instead of GLX, which in turn allows it to compile for Wayland. This can be done at the same time as using GLX on X11 for an application that works with either of the two. Right now, you can only use framebuffer surfaces on X11, but Skia supports doing so for Wayland just fine as long as you enable it. I added the `x11` feature as a default feature, but it doesn't actually do anything unless `gl` is also enabled. This means that anyone doing `skia-safe = { version = "...", features = ["gl"] }` won't notice a difference. This is _technically_ a breaking change, though, as anyone supplying `features = ["gl"], default_features = false` and then using the Skia functions to create a surface from a framebuffer will find that their code no longer correctly links. There's no reason anyone would do that since `skia-safe` has no default features right now, but it's worth keeping in mind.


pragmatrix:
Many thanks for this, I've rebased this branch on master, added some documentation and enabled the `egl`,`x11`,`wayland` features on the CI so that they are supported in the future. Also the image we use to build on the CI had to be updated to support EGL (specifically installing the `libgles2-mesa-dev` package).

I've removed the default feature `x11` from skia-safe, because it adds another library dependency, which I am not sure is available on all Linux distributions, and it also seems very confusing to have this enabled by default when you come from any another platform than Linux.

PR #436, opened 2020-12-11
Fix line endings

Vurich:
Many files in this repo had windows line endings, which are considered non-idiomatic by both Git and Rust. This fixes that issue.


pragmatrix:
This is embarrassing on many levels. First, I was aware of this problem, secondly, I've always though that text=auto in .gitattributes would prevent that, and thirdly, I've never recognized that that happened. Thanks for the PR. Rebased, and I've also added an .editorconfig file.

PR #437, opened 2020-12-11
Add docs

eira-fransham:
This builds on top of #436, so the diff will be hard to read without whitespace changes hidden (click the cog icon in GitHub's diff view).

`Handle` and `RCHandle` were not being exported from the top-level of `skia-safe`, which meant that viewing the docs would hide the vast majority of functions - as they are implemented mostly on `Handle<T>` or `RCHandle<T>` instead of `T`. While I think a nicer design would be to use newtypes instead of type aliases in order to improve docs and error messages, this is a nice middle-ground that doesn't break backwards-compatibility. This PR also puts many test functions into test modules, which I can split into a separate PR but I figured that I'd include it since it was already part of the same commit in my master branch anyway and I had enough PRs open already.

### DIFF
--- a/skia-safe/src/core/path.rs
+++ b/skia-safe/src/core/path.rs
@@ -16,24 +16,12 @@ pub use path_types::PathDirection as Direction;
 pub use path_types::PathFillType as FillType;
 
 pub use skia_bindings::SkPath_ArcSize as ArcSize;
-#[test]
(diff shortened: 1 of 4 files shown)


pragmatrix:
Closing, path.rs and surface.rs are documented in master for a while now.

Issue #356, opened 2020-07-24
Support Alpine / musl


arraypad:
At the moment skia-bindings/build_support/skia.rs targets GNU libc explicitly for Linux.

If I remove that restriction (or add `(_, "unknown", "linux", Some("musl"))`):
```diff --git a/skia-bindings/build_support/skia.rs b/skia-bindings/build_support/skia.rs
index 2133300..dfc1812 100644
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -372,7 +372,7 @@ impl BinariesConfiguration {
         let mut link_libraries = Vec::new();
 
         match target.as_strs() {
-            (_, "unknown", "linux", Some("gnu")) => {
+            (_, "unknown", "linux", _) => {
                 link_libraries.extend(vec!["stdc++", "fontconfig", "freetype"]);
                 if features.gl {
                     link_libraries.push("GL");
```

...and install ninja and gn via apk, and convince the build to use them rather than from downloaded deps:
```
RUN git clone --depth 1 --single-branch --branch m84-0.30.1 https://github.com/rust-skia/skia.git /skia
ENV SKIA_OFFLINE_SOURCE_DIR=/skia \
	SKIA_OFFLINE_NINJA_COMMAND=/usr/bin/ninja \
	SKIA_OFFLINE_GN_COMMAND=/usr/bin/gn
```

...and build with `-crt-static`:
```
ENV RUSTFLAGS="-C target-feature=-crt-static"
```

...then it all works ok!

It would be helpful for me, and I imagine for others, if rust-skia could support Alpine/musl out of the box.
I know there's a risk in conflating Alpine and musl, but it's handy in this case that Alpine has packages for gn and ninja. Maybe the branch matching the musl ABI is conditional on having those executables in the PATH already?

I can spend a bit more time to whittle down my Dockerfile to a minimal reproduction if that's helpful, let me know.

FWIW the size of my docker image using rust-skia went from 145MB based on debian:10-slim to 63MB using alpine:3.12.0

pragmatrix:
I'd really like to support more target platforms.

But assuming that you mean with "support", support in the sense that binaries are built and released, I am a bit hesitant to add more platform targets, specifically when they can not be built on the hosts we already support (I've tried to build a musl binary from a Ubuntu 20 distribution, but failed so far).

[...]

PR #447, opened 2020-12-25
Use gn and ninja to compile the rust-skia bindings

pragmatrix:
This PR uses gn and ninja to compile the rust-skia C++ sources:

Pros:

- No more differences between how we build them and Skia's build script builds its sources. Skia's build script was always better in configuring the build flags and aligning them was always a pain.
- No separate skia-bindings library anymore, the Rust wrappers are directly included in the Skia libraries.
- Build fails earlier while in development.
- Build is probably a bit faster, because the bindings can be built in parallel.
- Less warnings in the logs.

Cons:
- One more commit that is needed to be maintained on top of the Skia code base.
- Makes using a vanilla Skia or offline build even more complicated.

/cc @Vurich


jedahan:
Does this make it easier or harder to build against musl as seen in #356 ?

pragmatrix:
@jedahan I think it should not make a difference, because the Skia branch in use is already tainted (i.e. needs a number of commits on top of it to compile with), **but** this PR would introduce the first commit that would be essential, meaning that there are probably a number of simpler rust-skia configurations and platform targets that can be built with a vanilla Skia branch of the same milestone release by utilizing offline build support. On the other hand, the coupling between rust-skia and Skia is very strong anyways, which leads me to think that what this PR does is beneficial overall.  Despite of that, I am waiting for more feedback and considerations before I consider this PR a net positive. For now, it's just a first working draft of an idea.

pragmatrix:
Closing, even if compilation of the bindings conceptually a lot more difficult, the separation is probably worth it. I want to force as less patches as possible on top of Skia.

PR #450, opened 2020-12-29
fix: paragraph line through enum error

StringKe:
https://github.com/rust-skia/rust-skia/blob/0b0aefcca9e5d40ea408972999bd04423c0d9043/skia-safe/src/modules/paragraph/text_style.rs#L15


pragmatrix:
Nice catch, thank you a lot!

PR #479, opened 2021-02-27
allow use of system libraries over building them outright + add nix development shell files

superwhiskers:
# overview

this pr adds (and documents) a new environment variable, `SKIA_USE_SYSTEM_LIBRARIES`, which controls the use of system libraries instead of forcing them to be built alongside skia. it also adds nix development shell files for those who use the nix package manager and [nix-direnv](https://github.com/nix-community/nix-direnv)

# motivation

i wanted to package a project that made use of this project for the nix package manager, but i found that the existing `SKIA_OFFLINE_*` environment variables didn't provide enough control over the build process to make writing a package that builds depends upon rust-skia a reasonable/simple enough thing to do


superwhiskers:
(waiting to see if skia builds under the development shell)

superwhiskers:
there. i did some refactoring of the build system and removed the online/offline distinction, as the other two environment variables were removed. it is, however, still referred to as "offline mode" in the logs and in the argument to `build_support::skia::build`. i have also updated the flake to where it can now build skia

pragmatrix:
This looks great already.

> removed the online/offline distinction

... which seems like a good decision.

Can you explain to me (and perhaps also document) why you changed the path resolvement of Python?

I would also appreciate some pointers, a script, or a Dockerfile that explains how to build rust-skia for NixOS. I never did build anything for NixOS, but plan to add an additional pre-release step to validate some build configurations that are not covered by the CI to improve long term maintenance. And if you want to take on that responsibility specifically for NixOS, you may add yourself to the top level `README.md` under the maintainers section.

Also I am not sure if switching of the `git-sync-deps` will always work if the use of system libraries is enabled. There are dependencies that seem to be used for certain features like `d3d12allocator`. But I think we can fix this later if it gets an issue for a use case unforeseen.

superwhiskers:
> Can you explain to me (and perhaps also document) why you changed the path resolvement of Python?

that was me forgetting that the `gn` in `bin/gn` doesn't run on nixos at all, so i rewrote the entire proceduce thinking it would work. it acts the exact same with the exception of finding the full path to python. i can remove it if you like.

> Also I am not sure if not running git-sync-deps will always work if the use of system libraries is enabled. There are dependencies that seem to be used for certain features like d3d12allocator. But I think we can fix this later if it gets an issue for a use case unforeseen.

i'm pretty sure i set it to not run if that's enabled, see this code snippet on lines 617-630 of `skia-bindings/build_support/skia.rs`.

```rust
if !offline && !build.use_system_libraries {
    println!("Synchronizing Skia dependencies");
    prerequisites::resolve_dependencies();
    assert!(
        Command::new(python2)
            .arg("skia/tools/git-sync-deps")
            .stdout(Stdio::inherit())
            .stderr(Stdio::inherit())
            .status()
            .unwrap()
            .success(),
        "`skia/tools/git-sync-deps` failed"
    );
}
```

> I would also appreciate some pointers, a script, or a Dockerfile that explains how to build rust-skia for NixOS. I never did build anything for NixOS, but plan to add an additional pre-release step to validate some build configurations that are not covered by the CI to improve long term maintenance. And if you want to take on that responsibility specifically for NixOS, you may add yourself to the top level README.md under the maintainers section.

i'm not 100% sure either as i'm only doing this so i can progress in my packaging of https://github.com/Kethku/neovide, the same thing that #340--the issue that caused the offline bulid stuff to be introduced to begin with--was attempting. if you like, i can tell you what i ended up doing in order to get neovide working once i have that finished.

superwhiskers:
regarding packaging for nixos, you wouldn't really "package" it, it would only be pulled as a dependency when building other crates. however, writing some form of tests for it might be useful

pragmatrix:
> i can remove it if you like.

Yes, please. 

> if you like, i can tell you what i ended up doing in order to get neovide working once i have that finished.

I would greatly appreciate that.

superwhiskers:
done

rust-toolchain
```diff
@@ -0,0 +1 @@
+stable
```
pragmatrix, line 1: I think we should not require a specific toolchain.

skia-bindings/README.md
```diff
@@ -34,15 +34,27 @@ Whenever a new version of `rust-skia` is built from the `release` branch on our
...
-| `SKIA_OFFLINE_GN_COMMAND`    | The `gn` command to run. Either an absolute path or a path relative to Skia's source directory. | `bin/gn`                           |
+| Variable             | Description                                                                                                | Default                                    |
+| -------------------- | ---------------------------------------------------------------------------------------------------------- | ------------------------------------------ |
+| `SKIA_NINJA_COMMAND` | The `ninja` command to run. It can be either a command name or an absolute path.                           | `ninja` by default, `ninja.exe` on Windows |
+| `SKIA_GN_COMMAND`    | The `gn` command to run. It can be either a command name or a path that starts at Skia's source directory. | `bin/gn`                                   |
+
+### Changing the Skia source directory
+
+In some cases, one may wish to provide an alternate Skia source directory, for whatever reason.
```
pragmatrix, line 49: Please remove the `,for whatever reason`, it somehow sounds picky / judgmental and is redundant.

pragmatrix (changes requested):
Thanks a lot. 

I have two small suggestions that should be addressed before we can merge this PR.

rust-toolchain
superwhiskers: this is only necerssary so that `rustup` can automatically determine the toolchain being used and automatically select one

superwhiskers: otherwise, i'd need to set a directory override within `rustup` or set a global toolchain

pragmatrix: I understand, but we do have beta builds running on the CI (not enabled right now, because of an unresolvable conflict with clippy). The build is configured with a `rustup default beta`, so how should we test the repository with beta or nightly builds, if the toolset is fixed in the repository?

superwhiskers: you would do a directory override in the ci, with `rustup override set <channel>`

superwhiskers:
any updates on my response?

pragmatrix:
@superwhiskers I don't see why the rust edition has to be constrained to stable for any build, so if you have any idea to configure the nix build without changing adding the `rust-toolchain`  file, I would appreciate that.

superwhiskers:
the primary reason it is this way is because nixpkgs-mozilla lacks flake support (a relatively new nix feature, and while i _can_ work around this, it's much easier to use rustup) and setting `rust-toolchain` ensures toolchain consistency for anyone using rustup. while i could use something like https://github.com/nix-community/fenix, it only has the nightly toolchain.

pragmatrix:
> regarding packaging for nixos, you wouldn't really "package" it, it would only be pulled as a dependency when building other crates.

But if rust-skia is only ever built as a dependency, isn't rust-toolchain actually needed, because - say in the case of neovide - the build would run from its directory and would build rust-skia as a dependency using the toolchain defined in the neovide repository / flake?

Also, is it perhaps possible to add the toolchain configuration to the `shellHook` in `flake.nix`?

I've managed to installed the nix package manager on macOS, but when I try to `nix build`, it outputs: `error: flake 'git+file:/rust-skia' does not provide attribute 'packages.x86_64-darwin.defaultPackage.x86_64-darwin', 'legacyPackages.x86_64-darwin.defaultPackage.x86_64-darwin' or 'defaultPackage.x86_64-darwin'`

Do I need to run nixos (for example in a Docker container) to build this?

superwhiskers:
> But if rust-skia is only ever built as a dependency, isn't rust-toolchain actually needed, because - say in the case of neovide - the build would run from its directory and would build rust-skia as a dependency using the toolchain defined in the neovide repository / flake?

no. this isn't for building rust-skia as a package, it's for developing it independent of other packages—it provides a shell environment that gives you everything you need to work with the source code, which includes rust.

> Do I need to run nixos (for example in a Docker container) to build this?

no. the error you are seeing is the result of the flake not providing a package, it provides a development environment. there are two ways you should be able to use it. one is by calling `nix develop`, but if you want it to be entirely automatic when you enter a directory, you can use https://github.com/nix-community/nix-direnv, which is what i prefer

pragmatrix:
@superwhiskers My macOS installation of the nixOS somehow picked up wrong home paths, so I tried this again with an ubuntu installation. Specifically setting up the environment with an additional shell command `rustup override set stable` (inserted after the `export CC...` lines) seems to work for me: the `nix develop` environment can be used with the stable toolchain while leaving the host's toolchain preference untouched. Is this a viable compromise, or did I miss something?

superwhiskers:
i suppose that's okay. my main reason for preferring that it be a file was to avoid inserting additional data into the configuration of rustup and also making it be selected by default on non-nix supported platforms

pragmatrix:
Can't push to your repository, I will make the changes discussed in master.

superwhiskers:
alright. thank you

PhilTaken:
@superwhiskers could you maybe share how you managed to package neovide? 

I'm currently trying to do it myself, unsuccessfully.

superwhiskers:
apologies for the late-ish response. i never got around to fully packaging it (i'm waiting on neovide to update their version of rust-skia,) but i do have a wip expression that'll get updated to work the second that the version is updated [here](https://github.com/superwhiskers/nya/blob/master/packages/neovide.nix)

and likely, once it works, i'll make an attempt at getting it into nixpkgs

PhilTaken:
Ah gotcha. I almost got the release binary to work, its just crashing instantly with "could not create a surface" but that might be hardware-related... 

If its not too much of a hassle, could you please notify me once it works? :)

I really hope I'm not breaking any rules talking about neovide in this PR

superwhiskers:
if you want, you can fire me a message on telegram using the link in my github profile or use my email (which is also at my github profile,) as i'd rather not keep discussing this in an unrelated repository / pull request. aside from that, i'll try and pr neovide with rust-skia updated to speed things along.

PR #486, opened 2021-03-14
Remove the need to explicitly initialize ICU on Windows

### LINKED ISSUE #557: icu::init panics on Windows unless run as Administrator
When the `textlayout` feature is enabled on Windows, my understanding is that client code should call `icu::init()` before attempting to do any typesetting so it can extract the `icudtl.dat` data and write it to disk. Skia itself seems to [strongly prefer](https://github.com/google/skia/blob/eb0195e5b4d5d61dfa97483c15e519b56c6a1a98/third_party/icu/SkLoadICU.cpp#L103) that it live in the same directory as the current executable (though it will also look in the library’s directory after printing a somewhat misleading error message suggesting it failed).

The current `icu::init` behavior of ensuring `icudtl.dat` is in the executable dir (and [panicking](https://github.com/rust-skia/rust-skia/blob/71c4728704abb4394077d5ea1b9faaf5c2ad5797/skia-bindings/src/icu.rs#L14-L15) if it's unable to create it) is causing [problems](https://github.com/samizdatco/skia-canvas/issues/35) for my Node library since in this case the executable is Node itself which lives in "Program Files" and thus requires Administrator access to modify.

It seems like the ideal solution would be to merge pull request #486 and dispense with the external file altogether. Is there anything currently blocking this that I could help out with? Alternatively, could `icu::init` check for the data file in the library directory as well before exiting?

pragmatrix:
This PR attempts to completely remove the dependency on the `icudtl.dat` file in the current directory on Windows and the invocation of the `icu::init()` function in skia-safe.

Closes #557

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 
 Skia Submodule Status: chrome/m93 ([upstream changes][skia-upstream], [our changes][skia-ours]).
 
-[skia-upstream]: https://github.com/rust-skia/skia/compare/m93-0.41.1...google:chrome/m93
-[skia-ours]: https://github.com/google/skia/compare/chrome/m93...rust-skia:m93-0.41.1
+[skia-upstream]: https://github.com/rust-skia/skia/compare/m93-0.42.0...google:chrome/m93
+[skia-ours]: https://github.com/google/skia/compare/chrome/m93...rust-skia:m93-0.42.0
 
 ## Goals
 
--- a/patch/all-features-macos.patch
+++ b/patch/all-features-macos.patch
@@ -6,8 +6,8 @@ index 078c8119..2d5f02b0 100644
  depot_tools = "fade894"
  
  [features]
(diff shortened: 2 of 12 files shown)


pragmatrix:
@samizdatco This looks good to me and supports the legacy behavior, too. Please take a look if you don't mind.

samizdatco:
This looks terrific! Thanks so much for revisiting it.

samizdatco:
I finally got a chance to try building against the 0.42.0 release that incorporated these changes and have found that when using Paragraph-based typesetting on Windows, no text is rendered at all (and no errors are reported along the way).

Nothing jumps out at me in these commits that would cause this though. Is it possible that something else changed upstream with milestone 93 that could be behind the regression? I haven't tried testing the `textlayout` module on WIndows with pure rust code yet (just through my javascript bridge) but will report back if I can replicate the bug there as well...

PR #487, opened 2021-03-14
Support access to the Typeface's font data

### LINKED ISSUE #485: Typeface OpenStream
I'm looking at implementing pure rust text shaping via RustyBuzz, but RustyBuzz needs to be able to parse a font in order to shape it. Skia appears to have a way to get the font data from a given typeface via https://api.skia.org/classSkTypeface.html#a48befdd7cb646d78baccad4937c832da but it seems like that hasn't been implemented, and a cursory look into the bindings seems like its not their either.

Is there some way to get at the actual font path or font data that I've missed? If not are there any pointers for interacting with skia's streams so that I could go implement the missing functions?

I know there are bindings for Harfbuzz in skia-safe with the text shaping feature, but I was looking into options that didn't require copying a `icudtl.dat` file to the output directory. Any tips?

pragmatrix:
Closes #485

### DIFF
--- a/skia-safe/src/core/typeface.rs
+++ b/skia-safe/src/core/typeface.rs
@@ -1,4 +1,4 @@
-use crate::interop::{MemoryStream, NativeStreamBase};
+use crate::interop::{MemoryStream, NativeStreamBase, StreamAsset};
 use crate::prelude::*;
 use crate::{font_arguments, interop, FontArguments};
 use crate::{font_parameters::VariationAxis, Data, FontStyle, GlyphId, Rect, Unichar};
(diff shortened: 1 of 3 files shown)


skia-safe/src/core/typeface.rs
```diff
@@ -237,7 +237,19 @@ impl RCHandle<SkTypeface> {
...
+    pub fn to_font_data(&self) -> Option<(Vec<u8>, usize)> {
+        let mut ttc_index = 0;
+        StreamAsset::from_ptr(unsafe { sb::C_SkTypeface_openStream(self.native(), &mut ttc_index) })
+            .and_then(|mut stream| {
+                let length = unsafe { sb::C_SkStreamAsset_getLength(stream.native()) };
+                let mut data = vec![0u8; length];
+                let stream = stream.native_mut().as_stream_mut();
+                let read =
+                    unsafe { sb::C_SkStream_read(stream, data.as_mut_ptr() as _, data.len()) };
+                (read == data.len()).if_true_some((data, ttc_index.try_into().unwrap()))
```
Kethku, line 249: Would the index here always be zero? Should it be passed in?

pragmatrix: From the Skia documentation:
```
     *  Return a stream for the contents of the font data, or NULL on failure.
     *  If ttcIndex is not null, it is set to the TrueTypeCollection index
     *  of this typeface within the stream, or 0 if the stream is not a
     *  collection.
     *  The caller is responsible for deleting the stream.
```
It is returned only and indicates the index of the Typeface inside a true type collection, otherwise zero.

Issue #507, opened 2021-04-22
Linking errors when having both rust-skia and mozjpeg dependencies


romanzes:
When running `cargo build` in my project that uses both `skia-safe` and `mozjpeg` crates, I'm getting a lot of errors on Ubuntu 18.4.5 (mac is fine though), all of which look similar to the following:
```
/home/roman/workspace/font-metrics-extractor/target/debug/deps/libskia_bindings-47e0a85be893853d.rlib(libjpeg.jsimd_none.o): In function `jsimd_ycc_rgb_convert':
          jsimd_none.c:(.text.jsimd_ycc_rgb_convert+0x0): multiple definition of `jsimd_ycc_rgb_convert'
          /home/roman/workspace/font-metrics-extractor/target/debug/deps/libmozjpeg_sys-25d217b8eeeac2cd.rlib(jsimd_none.o):/home/roman/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjpeg-sys-0.10.13/vendor/jsimd_none.c:70: first defined here
          collect2: error: ld returned 1 exit status
```
Here's an example of the minimal setup that reproduces this error: https://github.com/romanzes/font-metrics-extractor/blob/master/Cargo.toml
Having just `mozjpeg` or just `rust-skia` in the dependencies works fine, it's only when there are both the build would fail.
Side note: this started happening since 0.35.0 release, the previous releases didn't have this issue.

pragmatrix:
@romanzes Can you take a look at #512 and see if you can work around the linking problem as suggested in the PR.

PR #512, opened 2021-05-05
Allow jpeg-turbo as a -sys library

therealbnut:
Hey,

I was having some issues where I had two `-sys` libraries which were both linking `-ljpeg`, leading to duplicate symbol linker errors. I think that `rust-skia` normally detects the problem if `build.rs` has `cargo:rustc-link-lib=jpeg`, but I'm not sure.

Anyway, I think I fixed the issue and I think it's a nicer solution. It fixes it by adding a libjpeg sys crate, and allowing skia to use it instead. I think this is a good solution as it avoids unnecessary compilation, and allows you to interoperate with the same libjpeg library.

I think it'd be really nice if all the skia third-party dependencies in this repo were slowly moved over to a similar system.

Please let me know if you have any questions, or want any changes (I'm also happy if it's easier for you to make small tweaks to get it merge ready).

**note: I had some azure related issues, so I thought I'd check if this change is desired before trying to fix them**

Thank you!

### DIFF
--- a/skia-bindings/Cargo.toml
+++ b/skia-bindings/Cargo.toml
@@ -48,11 +48,15 @@ textlayout = []
 webp = ["webp-encode", "webp-decode"]
 webp-encode = []
 webp-decode = []
+# sys libraries
+use-system-jpeg-turbo = ["mozjpeg-sys"]
 # deprecated since 0.25.0
 svg = []
 shaper = ["textlayout"]
 
 [dependencies]
+
+mozjpeg-sys = { version = "0", features=["with_simd"], optional=true }
 
 [build-dependencies]
 cc = "1.0.37"
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -295,7 +295,10 @@ impl FinalBuildConfiguration {
                 cflags.extend(jpeg_sys_cflags.iter().map(|x| -> &str { x.as_ref() }));
                 args.push(("skia_use_system_libjpeg_turbo", yes()));
             } else {
-                args.push(("skia_use_system_libjpeg_turbo", no()));
+                args.push((
+                    "skia_use_system_libjpeg_turbo",
+                    yes_if(use_system_libraries),
+                ));
             }
 
             if let Some(opt_level) = &build.opt_level {
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -505,7 +505,9 @@ impl BinariesConfiguration {
         let feature_ids = features.ids();
 
         if features.text_layout {
-            additional_files.push(ICUDTL_DAT.into());
+            if target.is_windows() {
+                additional_files.push(ICUDTL_DAT.into());
+            }
             built_libraries.push(lib::SK_PARAGRAPH.into());
             built_libraries.push(lib::SK_SHAPER.into());
         }
--- a/skia-bindings/build_support/skia.rs
+++ b/skia-bindings/build_support/skia.rs
@@ -216,10 +216,6 @@ impl FinalBuildConfiguration {
                 ("skia_use_gl", yes_if(features.gl)),
                 ("skia_use_egl", yes_if(features.egl)),
                 ("skia_use_x11", yes_if(features.x11)),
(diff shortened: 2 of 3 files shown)


skia-bindings/build_support/skia.rs
```diff
@@ -692,7 +700,7 @@ pub fn build_skia(
     ninja_command: &Path,
 ) {
     let ninja_status = Command::new(ninja_command)
-        .args(&["-C", config.output_directory.to_str().unwrap()])
+        .args(&[":skia", "-C", config.output_directory.to_str().unwrap()])
```
therealbnut, line 703: Note: I did this so that it wouldn't build other executable (example) targets, and only build `libskia.a`. This avoids some other linker issues.

skia-bindings/Cargo.toml
```diff
@@ -48,12 +48,16 @@ textlayout = []
 webp = ["webp-encode", "webp-decode"]
 webp-encode = []
 webp-decode = []
+# sys libraries
+use-system-jpeg-turbo = ["mozjpeg-sys"]
```
therealbnut, line 52: `mozjpeg` is a fork that is ABI compatible with libjpeg, and can be configured to be the same, but it has a default configuration that leads to superior compression.

skia-bindings/build_support/skia.rs
```diff
@@ -691,7 +699,10 @@ pub fn build_skia(
     config: &BinariesConfiguration,
     ninja_command: &Path,
 ) {
+    // Libraries we explicitly want ninja to build.
+    let ninja_built_libraries = config.built_libraries.iter().filter(|x| *x != lib::SKIA_BINDINGS);
     let ninja_status = Command::new(ninja_command)
+        .args(ninja_built_libraries)
```
therealbnut, line 705: Explicitly ask ninja to build the targets we want. Otherwise we end up having linker errors for skia's example programs that we don't want anyway.

skia-bindings/build_support/skia.rs
```diff
@@ -691,7 +699,13 @@ pub fn build_skia(
     config: &BinariesConfiguration,
     ninja_command: &Path,
 ) {
+    // Libraries we explicitly want ninja to build.
+    let ninja_built_libraries = config
+        .built_libraries
+        .iter()
+        .filter(|x| *x != lib::SKIA_BINDINGS);
     let ninja_status = Command::new(ninja_command)
+        .args(ninja_built_libraries)
```
therealbnut, line 708: Explicitly ask ninja to build the targets we want. Otherwise we end up having linker errors for skia's example programs that we don't want anyway.

pragmatrix:
Because the nightly build went through, I suspect that the Azure problems are a side effect of giving ninja specific targets. I'll try later to see if we can find a solution.

BTW: I like the idea of running ninja with specific targets, because it could prevent us from applying 3d60f5a70c0de0634cfea5883b600b9857de2416 for every new Skia version.

And does this PR fix #507?

therealbnut:
Ah yes, it does fix that :)

therealbnut:
Nice, thank you it looks like that fixed the issue

therealbnut:
@pragmatrix it might be worth looking into before merging: I _think_ there may be an issue where `libjpeg` still gets linked.

I'm just working on trying to reproduce it, but it's possibly because of this (assuming `gn` leads to it being in there somehow):
https://github.com/rust-skia/skia/blob/chrome/m81/third_party/libjpeg-turbo/BUILD.gn#L14

If that's it I believe an easy way to fix this might be to introduce an empty library (no symbols) that `-ljpeg` will find first. It should work if we emit a file `libjpeg.a` with the contents `b"!<arch>"` and a higher search path precedence, although I'm not sure where to modify `build_support/skia.rs` to do that yet.

pragmatrix:
@therealbnut I will merge this now. Thank you for the PR. Also, I've fixed one (presumably accidental) change in the default behavior when then env variable `SKIA_USE_SYSTEM_LIBRARIES` was set.

therealbnut:
Good catch, thanks for your help!

therealbnut:
Also congratulations on 2^9 PRs 🎉

PR #525, opened 2021-06-11
Add a `binary-cache` feature

therealbnut:
I've been trying to integrate rust-skia into a build system which tightly controls system dependencies (nix, bazel), and I've found it's much easier to get it working if we can disable the binary cache.

Unfortunately `FORCE_SKIA_BUILD` only helps a little as it still ends up getting a bunch of dependencies (`"ureq", "flate2", "tar"`) which require maintenance, and have build issues.

This PR introduces a feature flag so that we can remove everything related to the binary cache, if it's not needed.

Everything related to the binary-cache has moved into `build_support/binary_cache`. This PR is a pure refactor, there are no functional changes, code hasn't been added or removed, only moved.

Edit: perhaps `remote_cache` is a better name than `binary_cache`. I was trying to stay consistent with the existing `binaries.rs` naming, but it might have changed a bit since that was first named.


therealbnut:
I'm not sure if you want to hold off on this PR. It's possible I'm misunderstanding something about `[build-dependencies]`, but it still seems to be adding the optional dependencies to my project's `Cargo.lock`.

therealbnut:
I don't think it's a bad refactor either way, I think it at least saves building that extra code.

pragmatrix:
So far I think it's great! And you are right, the name of the feature should probably changed, because it is about downloading Skia as an external dependency (binaries or sources) instead of expecting it in the source tree. 

I'll take a closer look this weekend.

therealbnut:
> So far I think it's great!

Thanks!

> it still seems to be adding the optional dependencies to my project's Cargo.lock

I worked it out, I was including `skia-safe` in multiple crates in a workspace. I fixed it by adding `default-features = false` to all of those places `skia-safe` is defined as a dependency, which makes sense in retrospect as it's additive, and adds that set of features. There's no issues with this PR in this regard.

pragmatrix:
This can can be merged as it is. Thank you a lot.

The problem with the storage space limits on Azure is getting addressed in https://github.com/pragmatrix/rust-skia/pull/8

therealbnut:
@pragmatrix Sorry - I forgot to say earlier, thank you for the quick review and merge!

Issue #453, opened 2021-01-05
Reuse existing Skia binaries if possible


JackRedstonia:
Currently, it seems that every time the `skia-bindings` crate is recompiled (for example when compiling after upgrading the Rust toolchain), it redownloads Skia, as `sudo iptraf` reports traffic when `cargo` reaches `skia-bindings (build)`.

Normally this is not that much of an issue as wherever it's downloading from serves the files quickly enough, but since yesterday it has become excruiciatingly slow and even corruptions occurred.

As per a comment in issue #10 (https://github.com/rust-skia/rust-skia/issues/10#issuecomment-754653140), the build script can be improved by adding a check if the desired binaries are already downloaded.

pragmatrix:
@JackRedstonia You've mentioned that the download happens every time a compiler flag is changed. Can you provide one example that can be used to prove a working fix? I am asking because I assume that the output directory gets changed for even the smallest environment change with the consequence that downloaded archives must be cached somewhere else.

JackRedstonia:
@pragmatrix Something like this in `Cargo.toml` of a project that depends on `skia-bindings` for example:
```toml
[profile.release]
lto = "thin"
```
Commenting/uncommenting this results in a rerun of the build script it seems, which in turn redownloads the binaries, plus some other stuff from `codeload.github.com`.
I tried cloning the repository and swapping out all the URLs with `file://`s to local downloads, but it still seems to download *some*thing.
(local downloads were obtained through proxies, GitHub *always* chooses the worst file server to download from for my region, resulting in the build taking forever and then failing on a network error)

JackRedstonia:
Adding new dependencies to a project also rebuilds.

Issue #429, opened 2020-11-09
Potential simplification of offline building and caching


therealbnut:
Hey thanks for rust-skia I use it all the time!

I'm using [sccache](https://github.com/mozilla/sccache), which caches some rust build outputs. In case you haven't encountered it, `ccache` the offical [skia supported way](https://skia.org/user/build) to cache builds and `sccache` adds rust support. We've found that fetching the skia binary on our CI can be fairly flaky, but our `sccache` cache is hosted on with our CI so has much better reliability, and we're using it anyway.

Due to network flakiness we need to separate `cargo fetch` from `cargo build`. On our systems a fetch can be retried, but the build itself is intended to be deterministic.

To separate `fetch` and use `sccache` instead of your skia build caching I believe your recommended steps are:
 * disable fetching  with the environment variables in `build.rs`
 * special-case/duplicate the build.rs fetching of skia alongside our `cargo fetch` command
 * set the environment variables in `build.rs` to specify where we've downloaded skia

I imagine this will all work, but it seems like a lot of setup, only works with un-forked Skia binaries, and also requires us to special-case a bunch of stuff for fetching that adds knowledge of a transitive dependency several layers above our rust code.

---

**TL;DR**

I've had an idea to improve this, and wanted to get your thoughts on it. It might result in less fetching and caching code for you, and might mean you don't need to manually upload binaries with release. The idea is:

You could move `build.rs` (and its module) into `rust-skia/skia`, which becomes a crate (not submodule). Use cargo to do the fetching, as sccache to do the binary caching. You could support both if there's any reason to by putting it behind a `feature` config.

---

The advantage of this will be:
 * A crate means that `cargo fetch` will fetch it, and it will play well with future `rust` ecosystem changes
 * You get the binary caching logic "for free" with `sccache`
 * skia supports `sccache`, so changes to skia itself will be cached and building it will be faster
 * I haven't checked, but you might be able to reuse a Google official ccache repo for Skia if there's a public one
 * If `cargo` has more caching solutions in future using `cargo` in a more vanilla way is more likely to be compatible

The disadvantage (if it's not opt-in) will be:
 * You need to fetching the source even if you only ever need the cached build
 * To get a build binary shared with the community you might need to set up sccache storage
 * If you don't set up a community read-only cache, then people will need all the dependencies needed to build Skia
 * People that want the caching will have to install sccache

[...]

PR #527, opened 2021-06-23
Add support for using a locally built Skia library

therealbnut:
This PR does two main things:

 * adds a `build-from-source` feature
 * adds environment variables to support using a locally built skia library (`SKIA_BUILD_DEFINES` and `SKIA_LIBRARY_SEARCH_PATH`).

My intention is to use this to simplify the build process with external build systems (nix+bazel).

Most of the changes in this PR are moving code without changing functionality, primarily `skia-bindings/build_support/skia.rs` has been split into multiple parts:

 * `build_support/binaries_config.rs`
 * `build_support/bind_skia.rs`
 * `build_support/build_skia`
 * `build_support/features.rs`

I've tested this on mac under these conditions:

```bash
cargo build
```

```bash
cargo test --all --features=textlayout,use-system-jpeg-turbo,binary-cache,build-from-source
```

```bash
SKIA_SOURCE_DIR='/path/to/skia' SKIA_LIBRARY_SEARCH_PATH='/path/to/skia/libs' \
SKIA_BUILD_DEFINES="-DSK_HAS_ANDROID_CODEC -DNDEBUG -DSK_SUPPORT_GPU=0 -DSK_ASSUME_GL=1 -DSK_ENABLE_API_AVAILABLE -DSK_GAMMA_APPLY_TO_A8 -DSKIA_IMPLEMENTATION=1 -DSK_SUPPORT_PDF -DSK_CODEC_DECODES_JPEG -DSK_ENCODE_JPEG -DSK_USE_LIBGIFCODEC -DSK_CODEC_DECODES_PNG -DSK_ENCODE_PNG -DSK_XML" \
cargo build -v -p skia-safe --no-default-features
```

I've also added an output file `ninja_defines.txt`, which makes getting values for `SKIA_BUILD_DEFINES` easier.

This PR doesn't address any other open issues I've found, but it might make it easier to do #10 and #453. It also means I'm happy to close an old issue of mine #429.


therealbnut:
Hmm, I’ll have a look into the build errors tomorrow (late at night for me at the moment)

therealbnut:
I'm trying to reproduce this now, running this on MacOS 10.15.7:
```
cargo clean
cargo build -p skia-safe --all-targets --release --features "gl,metal,textlayout,vulkan,webp" --target x86_64-apple-darwin
```
I even tried entirely removing `target` first...
```
rm -rf ./target/
```
I'm not sure how to get better output from `gn`. Although as that's what's failing I suspect it's something to do with `skia/tools/git-sync-deps`. My next steps are to verify that it is being called in the same way as it was before this PR - although I'm not sure why it would work locally, unless there's some special casing for CI.

therealbnut:
@pragmatrix should be ready for review now!

The last several commits are fixing clippy issues that seemed to be on master, I'm not sure if my changes are wanted - if you like I can split them out into a separate PR we can merge first.

To review the rest of this PR I've left some notes in the PR description, although if you wanted to follow the first few commit one at a time it should guide through you through what changed, and allow you to skip over a majority of the changes. I had one mistake where I passed the wrong boolean (to indicate offline builds), but that's fixed now.

I think it would be good to add a test for this new functionality, but I didn't want to mess with your CI stuff.

The way I was thinking of doing this is something like this:

 * build (or copy from another build) the skia libraries (`out/*.a`) and `ninja_defines.txt` into a temporary directory
 * set `SKIA_SOURCE_DIR` to `skia-bindings/skia`
 * set `SKIA_BUILD_DEFINES` from `ninja_defines.txt`
 * set `SKIA_LIBRARY_SEARCH_PATH` to your temporary directory
 * run `cargo build -v -p skia-safe --no-default-features` with whichever features you want, and run whichever tests you want
 
The `skia-bindings` target builds pretty quickly with these settings, so it shouldn't add much overhead, especially if you're doing it with libraries from an existing build.

pragmatrix:
Skipping over the changes, I had a few thoughts:

- I like the separation of the Skia build steps into separate files, this was long overdue.
- The Clippy changes should be separated, I'll prepare a PR against master, and will rebase this one. I hope you don't mind.
- Can we avoid the additional feature flag? This time, there seem no changes in the `[dependencies]` and the additional code to compile is negligible. I suspect that we can make the change in behavior base on the existence of `SKIA_LIBRARY_SEARCH_PATH`?.

> I think it would be good to add a test for this new functionality, but I didn't want to mess with your CI stuff.

We can take care of that after the merge. I think an additional workflow that runs supplementary builds for verifying certain configurations should be added.

Also some rename suggestions: 
- `build_skia/` -> `skia/` (because it's already under `build_support/`).
- `ninja_defines.txt` -> `skia-defines.txt` (contents are related more to the Skia library than to build system and text files should probably not follow Rust conventions, and I think `-` feels a bit more natural to type)
- `bind_skia.rs` -> `skia_bindgen.rs`

therealbnut:
@pragmatrix thanks for the feedback, that all sounds good!

I had the feature flag early on because I was expecting more concrete dependencies but that simplifies things.

therealbnut:
As a side note: I liked the symmetry of build and bind as separate "features". Perhaps in a future PR the bind things can be moved out of build_support as its own module.

therealbnut:
@pragmatrix those changes should be in now - I'm just running locally `cargo +beta clippy` but I don't expect any issues

therealbnut:
@pragmatrix clippy was all good locally - I think the only outstanding change is rebasing on your clippy PR, I'm happy if you want to do that :)

therealbnut:
Thanks for the review and rebase :)

PR #532, opened 2021-06-26
Support local builds with the textlayout feature enabled.

pragmatrix:
This PR attempts to support local builds (#527) with the `textlayout` feature enabled on Linux and Window.

@therealbnut We could, instead of just adding `SKIA_LIBRARY_SEARCH_PATH` to cargo's library search path, copy all the library and additional files to the output directory similar to what the `binary-cache` does that after downloading the binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,9 +98,9 @@ workflows:
 #
 # https://github.com/rust-skia/rust-skia/pull/527
 
-local-build-features=gl,vulkan,webp
+local-build-features=gl,vulkan,webp,textlayout
 
-.PHONY: test-local-build build-local-build prepare-local-build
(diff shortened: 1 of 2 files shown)


therealbnut:
> copy all the library and additional files to the output directory

I’m not sure if I’m fully understanding - but can you know that output directory’s path ahead of time? I’m happy if you can do it deterministically (I don’t see any issues with the PR at the moment).

I’d be happy instead if we specified a data path for additional files in another environment variable. Or combined data and libraries into a single path.

I’m considering system installs of libraries where the files might be in multiple locations, but that can be solved with symlinks.

If any of the libraries end up being dynamically linked in future it might be an issue, but that’s a future problem.

pragmatrix:
Closed, covered in #542

PR #542, opened 2021-07-09
Copy library binaries in local build instead of using the search path…

pragmatrix:
Instead of just adding the search path set in local builds to cargo, this PR copies the libraries to the output directory. The intention here is to be sure early on that all the files to link a binary are existing.

This PR is meant to supersede #532

### DIFF
--- a/skia-bindings/build_support/binaries_config.rs
+++ b/skia-bindings/build_support/binaries_config.rs
@@ -27,8 +27,9 @@ pub struct BinariesConfiguration {
     /// The static Skia libraries built by ninja that dependent projects need to link with.
     pub ninja_built_libraries: Vec<String>,
 
-    /// Other static Skia libraries skia-bindings provides that dependent projects need to link with.
-    pub other_built_libraries: Vec<String>,
+    /// Static libraries that are generated in the binding process dependent projects need to link
+    /// with.
+    pub binding_libraries: Vec<String>,
 
     /// Additional files relative to the output_directory
     /// that are needed to build dependent projects.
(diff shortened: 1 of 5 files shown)


pragmatrix:
@therealbnut Would you mind testing if this works for your local builds? Compared with `master` this branch copies all Skia libraries (except `libskia-bindings.a`) to the output directory. If this does not cause too much trouble, I'd rather would prefer it, because it is more consistent in terms that there are no _external_ references needed in the downstream linking processes and that the availability of the libraries is checked earlier up front.

skia-bindings/build_support/binaries_config.rs
```diff
@@ -146,7 +154,41 @@ impl BinariesConfiguration {
...
+        to_dir: &Path,
+        copy_bindings_libraries: bool,
+    ) -> io::Result<()> {
+        fs::create_dir_all(&to_dir)?;
+
+        let target = cargo::target();
+
+        for lib in self.built_libraries(copy_bindings_libraries) {
+            let filename = &target.library_to_filename(lib);
+            fs::copy(from_dir.join(filename), to_dir.join(filename))?;
```
therealbnut, line 185: There’s a lot of copying which might be slow and use disk space - I wonder if it’s better to symlink if available

(ie. `std::os::unix::fs::symlink` or `std::os::windows::fs::symlink_file`)

therealbnut (commented):
@pragmatrix Thanks for checking!

If I understand correctly I’d still specify where to find the libraries the same way - but they are copied during the build to make linking more hermetic?

I can’t see any reason that shouldn’t work - I can test later today if I get time, but I don’t think you need to wait.

The only thing I think would be good to improve is how this will lead to multiple copies of the files on disk.

skia-bindings/build_support/binaries_config.rs
pragmatrix: I suggested to take care of that at a later time if this gets to be an issue in some constrained build environments.

therealbnut: Sounds good to me

PR #553, opened 2021-08-03
Simple fix that replaces an array with the `vec![]` macro. 

colbyn:
Simple fix that replaces an array with the `vec![]` macro (to get Rust analyzer to chill). 

Rust analyzer keeps complaining that `libs.extend(["framework=MobileCoreServices", "framework=UIKit"]);` is not an iterator. 

Replaced such with `libs.extend(vec!["framework=MobileCoreServices", "framework=UIKit"]);`.

Thanks! 
Colbyn

### DIFF
--- a/skia-bindings/build_support/ios.rs
+++ b/skia-bindings/build_support/ios.rs
@@ -93,7 +93,7 @@ pub(crate) fn link_libraries(abi: Option<&str>, features: &Features) -> Vec<&'st
     ];
 
     if abi != Some("macabi") {
-        libs.extend(["framework=MobileCoreServices", "framework=UIKit"]);
+        libs.extend(vec!["framework=MobileCoreServices", "framework=UIKit"]);
     }
 
     if features.metal {


pragmatrix:
Is this rust-analyzer complaining or an older Rust version? I think that this is caused if a Rust version before version 1.53 is used to run `cargo check` or `cargo clippy` (`IntoIterator` for arrays is implemented since version 1.53, see also: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend-1).

colbyn:
@pragmatrix No, it builds just fine. I don't know what is wrong with rust-analyzer. The fix was simple, so I just thought I'd patch the library. But if you'd rather not merge it, I understand.

pragmatrix:
Merging, but I just wonder why this happens with your rust-analyzer installation and not mine. Ironically, I expect clippy to complain about this soon :).

colbyn:
Oh interesting, perhaps it's because I'm on nightly and you're on the weekly release, or somethings screwed up with my setup. 🤔

Issue #545, opened 2021-07-21
Add bindings for SkSVGDOM


samizdatco:
Would it be possible to add bindings for the [`SkSVGDOM`](https://github.com/google/skia/blob/main/modules/svg/include/SkSVGDOM.h) class? Being able to parse SVG files and draw them to the canvas would be a terrific enhancement. 

See also: [SvgTool.cpp](https://github.com/google/skia/blob/main/modules/svg/utils/SvgTool.cpp) for a minimal use case.

pragmatrix:
> Being able to parse SVG files and draw them to the canvas would be a terrific enhancement.

Agreed, and as I can see from the Network graph, @triniwiz already attempted to extend the bindings, may be you can collaborate?

pragmatrix:
@triniwiz Is it possible to draw SVG files with that branch? We don't have to implement the full SkSVGDOM API (Node, etc.) for a first MVP, rendering should cover most use cases I guess.

triniwiz:
Yes it works fine for my usage passing a string or path from Android/iOS to render . One more thing I know some are a bit picky with it comes to adding dependencies since I added reqwest. It was added from my understanding that embedders need to handle those asset loading e.g images , fonts(did not tackle this one) ..etc

pragmatrix:
Bindings for SkSVGDom are available for a long time now, closing, see #681.

PR #569, opened 2021-10-13
feat: initial svg rendering

triniwiz:
Related to #545 .

I'm sharing what I wrapped up to work with [NativeScript](https://github.com/NativeScript/canvas/tree/master/packages/canvas/src-native/canvas-native) to render SVG based on [this fork](https://github.com/Vurich/rust-skia) .


pragmatrix:
Great, CI says OK on all platforms now, things that came to me mind to finish this:

Required:

- [x] Replaced reqwest by ureq to remove the openssl-sys dependency.
- [x] Move `svg::Dom` to `skia-safe/src/modules`
- [ ] Take a closer look at the `NativeDrop` implementations. Some are not needed, some may leak memory.
- [ ] Remove `Svg` prefix from the types, they are already in a module named `svg`.
- [ ] Add tests for RustStram
- [ ] Add tests for `svg::Dom`.
- [ ] Make the ureq asset downloader more robust? Add a test for download errors.
- [ ] Merge imports.
- [ ] Clarify public API surface (would prefer `crate::svg::Dom`).
- [ ] Add example rendering to the `skia-org` program.
- [ ] Update various README.md (don't forget notable contributions).

Optional:
- [ ] Consider providing `RustStream` in all builds.
- [ ] Add a render pass of all existing images to SVG and back to an image.

ctrlcctrlv:
Woah, if I am understanding this code right, it uses liberally `skia::Canvas`—does this mean that I'd be able to let users take SVG "screenshots" of MFEKglif? As in, this but in SVG format?

image

MFEKglif's use of rust-skia is extremely extensive. We use almost all parts…PathOps, all the math stuff, all the bounds stuff, the path effects, etc. And, even though you see imgui there, actually [we are drawing imgui through rust-skia](https://github.com/MFEK/imgui-skia-renderer.rlib/).

This is really cool.

pragmatrix:
@ctrlcctrlv This should already be possible with rendering to an `skia-safe::svg::Canvas` (which implements `Deref` and `DerefMut` on `skia-safe::Canvas`. This PR is about rendering an SVG file to a Canvas (the other way around).

ctrlcctrlv:
Ohh okay. Well, both are useful to me—it'd be an easy way to let users import SVG files for example. Right now I maintain https://github.com/MFEK/glif2svg and there's a fonttools script called svg2glif.py, but both those are quite clumsy compared to literally just allowing user to open SVG file. Very neat :sunglasses:

pragmatrix:
@triniwiz I am trying to figure out why this base64 option in `handle_load()` is needed. Do you have an example. It looks to me that this is something of a hack to provide resources data through base64 encoded path strings. I've tried to load the following SVG:

```svg
        <svg version="1.1"
        xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
        width="128" height="128">
        <image width="128" height="128" transform="rotate(45)" transform-origin="64 64"
            xlink:href="https://www.rust-lang.org/logos/rust-logo-128x128.png"/>
        </svg>
```

and it lands in `handle_load()` with a `""` resource path, which causes `is_base64` to be set to `true`.

Also I am inclined to fully support the `skresources::ResourceProvider`, this way we don't have to add a dependency to ureq / reqwest in skia-safe and users can fully control the resource requests.

triniwiz:
Oh no, I could swore the resource path was only empty for base64 images 🤔 .

triniwiz:
Using your example I get the following logged.

```bash
resource_path "https://www.rust-lang.org/logos"
resource_name "rust-logo-128x128.png"
is_base64 false

```
 It still failed for me but after some digging I found on my end the issue was caused by me using `v.text()` with the reqwest so `v.bytes()` fixes it for me

pragmatrix:
This seems to be platform dependent, I've just tried on macOS, and there the URL gets split up as you say. On Windows the URL is in `resource_name`.

```
resource_path: ""
resource_name: "https://www.rust-lang.org/logos/rust-logo-128x128.png"
```

I've tried then to use `https://www.rust-lang.org\logos/rust-logo-128x128.png` instead, and on Windows it splits up like so:

```
resource_path: "https://www.rust-lang.org"
resource_name: "logos/rust-logo-128x128.png"
```

which means that on Windows only `\` is considered a separator for URL references, which is obviously wrong.

triniwiz:
Oh sorry I've haven't tried this on windows as it was out of my scope 😕

mpajkowski:
Hello,

what tasks need to be done to merge this feature?

triniwiz:
@pragmatrix Anything else needed for this ?

pragmatrix:
Closing in favor of #681

PR #575, opened 2021-11-01
Add feature embed-freetype

### LINKED ISSUE #570: FreeType 2.7+ required as of release 0.43.0?
It seems that a more recent version of FreeType was used for the last round of precompiled binaries which includes a call to `FT_Set_Default_Properties`. This function isn't present in versions of FreeType prior to 2.7 and is causing undefined symbol errors when run on older linux systems (in particular Debian Stretch which is still quite popular for use in Docker containers).

I'm unsure whether anything can/should be done to try to preserve backward compatibility or if this is under the control of `skia/third_party/freetype2` and we should adjust the system requirements accordingly. What do you think?

pragmatrix:
Closes #570

### DIFF
--- a/docker/linux/debian-stretch/Dockerfile
+++ b/docker/linux/debian-stretch/Dockerfile
@@ -1,0 +1,38 @@
+FROM debian:stretch-backports
+
+RUN apt-get update
+
+# Preinstall tzdata, so that it does not when installed as a transitive dependency later.
+ENV TZ=Europe/Berlin
+RUN DEBIAN_FRONTEND=noninteractive apt-get install tzdata
+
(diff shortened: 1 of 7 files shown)


ctrlcctrlv:
Awesome, thanks for this one, anything I can do to make MFEK (MFEK{glif,stroke,metadata…}) more portable is a win :-) :+1: 

I'm commenting here also so I remember to check in the morning if this branch breaks my `imgui0.8` build, since that _also_ adds support for FreeType. Hopefully the Cargo.toml's are compatible, they usually are, but if not things get nasty because Rust can't double link a C lib like it can a native lib.

PR #577, opened 2021-11-07
Update NDK to r23b

pragmatrix:
Once the containers are updated, we can try to build rust-skia with the latest Android NDK r23b.

**Update:** Reverted the containers. This is not as easy as expected, the Android NDK has changed significantly:

- [ ] Use llvm-ar instead of gnu utils ar.
- [ ] features.h not found (wrong -sysroot?)
- [ ] `-lgcc` (`libgcc.a`) not found (who refers to it?, suspicion is the `cc` crate).


pragmatrix:
Closing in favor of #704

PR #595, opened 2021-12-12
Use python 3

pragmatrix:
Experimental PR to see if python 3 can now be used to build Skia on all platforms.


pragmatrix:
I think we should take the risk and move to Python3 _now_, on Monterey it's not installed anymore. If this does not work well in practice, we'll prefer 2, but fall back to 3 in the future. So far I've seen problems only with msys on Windows.

PR #613, opened 2022-02-07
 Document `PathMeasure` API weirdness (inherited from Skia C++) (closes #600)

### LINKED ISSUE #600: `PathMeasure::length` API is confusing and poorly documented
Because a `PathMeasure` is formed from a path, it's not immediately clear that calling `length()` on the result will just return the length of the first contour.

Of course, this is also true of the C++ API, I do recognize. But, I think that `rust-skia` should consider at least documenting that you must e.g.:

```rust
    let mut measure = skia::PathMeasure::from_path(&skp, false, None);
    let mut slen = measure.length();
    while measure.next_contour() {
        slen += measure.length();
    }   
```

I had a bug related to this and it was quite frustrating to figure out.

ctrlcctrlv:
Sorry took me a while to PR you. I planned to shortly after opening my issue.

Closes #600.

### DIFF
--- a/skia-safe/src/core/path_measure.rs
+++ b/skia-safe/src/core/path_measure.rs
@@ -41,6 +41,22 @@ impl fmt::Debug for PathMeasure {
     }
 }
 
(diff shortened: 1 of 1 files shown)


pragmatrix:
Thank you a lot @ctrlcctrlv

PR #641, opened 2022-05-25
Fix homepage link for skia-bindings

marc2332:
Just a quick fix changing the homepage link in Cargo.toml from https://github.com/rust-skia/rust-skia/skia-bindings (doesn't exist) to https://github.com/rust-skia/rust-skia/tree/master/skia-bindings


pragmatrix:
Thanks, very appreciated!

PR #645, opened 2022-06-06
Support file:// paths for downloading binaries

### LINKED ISSUE #644: [Feature request] Suggest supporting getting prebuild packages from local fs
Due to the awful network environment (I'm in China), I need to download prebuild packages through VPN, but `ureq` seems won't recognize `https_proxy` env var. Thus I have to download them manually, starting a file server, then setting `SKIA_BINARIES_URL` to let skia download from it. It is convenient if there is a way to access these files directly.

pragmatrix:
Closes #644

### DIFF
--- a/skia-bindings/build_support/binary_cache/utils.rs
+++ b/skia-bindings/build_support/binary_cache/utils.rs
@@ -1,9 +1,25 @@
-use std::io;
-use std::io::Read;
+use std::{
+    fs,
+    io::{self, Error, ErrorKind, Read},
+    path::PathBuf,
+};
 
 /// Download a file from the given URL and return the data.
 pub fn download(url: impl AsRef<str>) -> io::Result<Vec<u8>> {
-    let resp = ureq::get(url.as_ref()).call();
+    let url = url.as_ref();
+
+    // `file` URL, empty hostname, absolute path
+    if let Some(file_path) = url.strip_prefix("file:///") {
+        return fs::read(PathBuf::from(format!("/{}", file_path)));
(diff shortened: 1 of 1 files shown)


skia-bindings/build_support/binary_cache/utils.rs
```diff
@@ -1,9 +1,25 @@
...
+    path::PathBuf,
+};
 
 /// Download a file from the given URL and return the data.
 pub fn download(url: impl AsRef<str>) -> io::Result<Vec<u8>> {
-    let resp = ureq::get(url.as_ref()).call();
+    let url = url.as_ref();
+
+    // `file` URL, empty hostname, absolute path
+    if let Some(file_path) = url.strip_prefix("file:///") {
```
Fancyflame, line 12: I think there is a compatibility problem on windows platform, since the absolute path on windows is something like `file://C:\path\to\file`. Requiring a file url starts with `file:///` then add extra `/` on the start of the url will change this example path to `/C:\path\to\file`. Obviously, it will break the path.
It is suggested to replace `url.strip_prefix("file:///")` with `url.strip_prefix("file://")` and replace `format!("/{}", file_path)`with `file_path`. Correct me if I'm misunderstood.

Fancyflame (changes requested):

skia-bindings/build_support/binary_cache/utils.rs
pragmatrix: Yes, that sounds about right.

PR #651, opened 2022-06-27
update flake.nix to include darwin m1 support

busti:
Allows for the project to be built on apple m1 machines using nixpkgs.  
Also bumps the versions in `flake.lock` and makes some minor changes to the names used in the flake since the naming conventions changed.

I have also taken the liberty of sorting the packages alphabetically as it makes it a little easier to find a specific package when the list gets too long.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


pragmatrix:
Thank you a lot, merging.

PR #655, opened 2022-07-08
Support cross-compilation for Windows arm64

syrel:
This pull request adds support for cross-compilation of `rust-skia` for `--target aarch64-pc-windows-msvc` with `crt-static`.

- When cross-compiling on Windows, `cfg!(target_feature = "crt-static")` always returns `false` even though `CARGO_CFG_TARGET_FEATURE` contains `crt-static` feature. 
- `target_cpu` should be specified similarly to linux and mac.

### DIFF
--- a/skia-bindings/build_support/cargo.rs
+++ b/skia-bindings/build_support/cargo.rs
@@ -114,7 +114,9 @@ pub fn target() -> Target {
 }
 
 pub fn target_crt_static() -> bool {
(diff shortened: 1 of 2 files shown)


pragmatrix:
Thank you @syrel. Nice catch with the `crt-static` flag not being set when cross-compiling. I've removed the `target_cpu` for `x86` builds to work around a build problem on 32 bit Windows targets.

PR #668, opened 2022-08-03
Add support for building against a Yocto SDK

tronical:
In a Yocto SDK environment the following relevant environment variables
are set:

 - SDKTARGETSYSROOT: points to the target sysroot.
 - CLANGCC/CLANGCXX: the command line to use for cross-compiling with
   clang to the target.
 - CC/CXX: the gcc command line for cross-compiling.

This patch introduces a dependency to these two variables, as preferred
sysroot and clang choices after the macOS specific SDKROOT and CC/CXX.

Another aspect to the Yocto build is that the target name diverges from
cargo.  For example what's aarch64-unknown-linux-gnu with cargo might be
aarch64-poky-linux (vendor set, no abi specified).

On a clang level this is dealt with by CLANGCC/CLANGCXX including
--target=aarch64-poky-linux for example.

Unfortunately the Skia build also adds --target to the command line,
based on Cargo's target. In the final command line when compiling Skia,
compiling bindings.cpp or generating the bindings, clang's attempt at
locating the correct gcc installation for libstdc++ headers fails,
because --target=aarch64-unknown-linux-gnu does not match the
compiled-in paths that clang would look for if the target was
aarch64-poky-linux (as per original command line).

In order to fix that, this patch also extracts any existing
--target=$foo from the build configuration and uses it throughout in
favor of cargo::target() if provided.

### DIFF
--- a/skia-bindings/build_support/skia/config.rs
+++ b/skia-bindings/build_support/skia/config.rs
@@ -43,10 +43,15 @@ impl BuildConfiguration {
             .or_else(|| cargo::env_var("CXX"))
             .unwrap_or_else(|| "clang++".to_string());
 
-        let target = cc.find("--target=").and_then(|target_option_offset| {
-            cc[(target_option_offset + "--target=".len())..]
+        // It's possible that the provided command line for the compiler already includes --target.
+        // We assume that it's most specific/appropriate, extract and use is. It might for example include
+        // a vendor infix, while cargo targets usually don't.
+        let target = cc.find("--target=").map(|target_option_offset| {
+            let target_tail = &cc[(target_option_offset + "--target=".len())..];
+            let target_str = target_tail
                 .split_once(' ')
-                .map(|(target_str, ..)| crate::build_support::cargo::parse_target(target_str))
+                .map_or(target_tail, |(target_str, ..)| target_str);
+            crate::build_support::cargo::parse_target(target_str)
         });
 
         BuildConfiguration {
--- a/skia-bindings/build_support/skia/config.rs
+++ b/skia-bindings/build_support/skia/config.rs
@@ -173,6 +173,8 @@ impl FinalBuildConfiguration {
             let mut cflags: Vec<String> = Vec::new();
             let mut asmflags: Vec<String> = Vec::new();
 
+            // SDKROOT is the environment variable used on macOS to specify the sysroot. SDKTARGETSYSROOT is the environment
+            // variable set in Yocto Linux SDKs when cross-compiling.
             if let Some(sysroot) =
                 cargo::env_var("SDKTARGETSYSROOT").or_else(|| cargo::env_var("SDKROOT"))
             {
--- a/skia-bindings/build_support/skia/config.rs
+++ b/skia-bindings/build_support/skia/config.rs
@@ -29,8 +29,8 @@ pub struct BuildConfiguration {
     /// C++ compiler to use
     cxx: String,
 
-    /// Optional target
-    pub target: Option<Target>,
+    /// The target (arch-vendor-os-abi)
+    pub target: Target,
 }
 
 /// Builds a Skia configuration from a Features set.
@@ -48,13 +48,16 @@ impl BuildConfiguration {
         // It's possible that the provided command line for the compiler already includes --target.
         // We assume that it's most specific/appropriate, extract and use is. It might for example include
         // a vendor infix, while cargo targets usually don't.
-        let target = cc.find("--target=").map(|target_option_offset| {
-            let target_tail = &cc[(target_option_offset + "--target=".len())..];
-            let target_str = target_tail
-                .split_once(' ')
-                .map_or(target_tail, |(target_str, ..)| target_str);
-            crate::build_support::cargo::parse_target(target_str)
-        });
+        let target = cc
+            .find("--target=")
+            .map(|target_option_offset| {
+                let target_tail = &cc[(target_option_offset + "--target=".len())..];
+                let target_str = target_tail
+                    .split_once(' ')
+                    .map_or(target_tail, |(target_str, ..)| target_str);
+                crate::build_support::cargo::parse_target(target_str)
+            })
+            .unwrap_or_else(cargo::target);
 
         BuildConfiguration {
             on_windows: cargo::host().is_windows(),
@@ -164,7 +167,7 @@ impl FinalBuildConfiguration {
             let mut use_expat = true;
 
             // target specific gn args.
-            let target = build.target.clone().unwrap_or_else(cargo::target);
+            let target = &build.target;
             let mut target_str = format!("--target={}", target);
             let mut set_target = true;
             let mut cflags: Vec<String> = Vec::new();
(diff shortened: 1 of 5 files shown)


tronical:
BTW, I'd be more than happy to make adjustments to the documentation or re-do parts of this approach. Just thought that instead of opening just an issue to complain it might be easier to discuss problem and solution right away with a patch :)

skia-bindings/build_support/skia/config.rs
```diff
@@ -27,19 +28,34 @@ pub struct BuildConfiguration {
...
 /// Builds a Skia configuration from a Features set.
 impl BuildConfiguration {
     pub fn from_features(features: features::Features, skia_debug: bool) -> Self {
+        let cc = cargo::env_var("CLANGCC")
+            .or_else(|| cargo::env_var("CC"))
+            .unwrap_or_else(|| "clang".to_string());
+        let target = cc.find("--target=").and_then(|target_option_offset| {
+            cc[(target_option_offset + "--target=".len())..]
+                .split_once(' ')
+                .map(|(target_str, ..)| crate::build_support::cargo::parse_target(target_str))
```
pragmatrix, line 45: `cargo::parse_target` should be enough to resolve the function.

skia-bindings/build_support/skia/config.rs
```diff
@@ -27,19 +28,34 @@ pub struct BuildConfiguration {
...
 
 /// Builds a Skia configuration from a Features set.
 impl BuildConfiguration {
     pub fn from_features(features: features::Features, skia_debug: bool) -> Self {
+        let cc = cargo::env_var("CLANGCC")
+            .or_else(|| cargo::env_var("CC"))
+            .unwrap_or_else(|| "clang".to_string());
+        let target = cc.find("--target=").and_then(|target_option_offset| {
+            cc[(target_option_offset + "--target=".len())..]
+                .split_once(' ')
```
pragmatrix, line 44: This assumes that a space follows `--target=XXX`, otherwise `None` is returned, and the target won't be extracted.

skia-bindings/build_support/skia_bindgen.rs
```diff
@@ -127,7 +131,7 @@ pub fn generate_bindings(build: &FinalBuildConfiguration, output_directory: &Pat
         .clang_args(&["-x", "c++"])
         .clang_arg("-v");
 
-    let target = cargo::target();
+    let target = target.unwrap_or_else(cargo::target);
```
pragmatrix, line 134: nit: I don't like this here, because it kind of defines a default "under the hood". It should be probably pulled up the call chain as far as possible so that `generate_bindings` only takes a `target: Target`, and not `target: Option<Target>`.

skia-bindings/build_support/skia_bindgen.rs
```diff
@@ -715,6 +718,8 @@ pub(crate) mod rewrite {
 
 pub use definitions::{Definition, Definitions};
 
+use super::cargo::Target;
```
pragmatrix, line 721: This `use` belongs to the top of the file. I am assuming that the IDE misplaced it because of the `pub use` before, which also should not belong here..

pragmatrix (commented):
Thank you a lot, I have some comments which are mostly optional to fix.

In addition to that:

- It would be great if there is a separate section in the README.md in skia-bindings that describes the environment variables that are expected to be set if a build with the Yocto SDK is intended (copying the text from the PRs description would be fine for a first draft). Perhaps a short description how to set up the SDK would be useful, too.
- If perfectionism calls, some of the build parameters, like sysroot may be pulled into `FinalBuildConfiguration`, too.
- Code comments would be helpful where environment variables override others.
- The most important point is to ensure that this does not break in the future. I am willing to maintain the changes, but can do so only with help and a GitHub workflow that builds a common / Slints feature configuration against the Yocto SDK (gl + textlayout?). This may be a separate PR, and I am willing to create a suitable workflow if you point me to instructions on how to set up a Yocto SDK build environment.

skia-bindings/build_support/skia/config.rs
```diff
@@ -27,19 +28,34 @@ pub struct BuildConfiguration {
...
             // `OPT_LEVEL` is set by Cargo itself.
             opt_level: cargo::env_var("OPT_LEVEL"),
             features,
             skia_debug,
-            cc: cargo::env_var("CC").unwrap_or_else(|| "clang".to_string()),
-            cxx: cargo::env_var("CXX").unwrap_or_else(|| "clang++".to_string()),
+            cc,
+            cxx: cargo::env_var("CLANGCXX")
+                .or_else(|| cargo::env_var("CXX"))
+                .unwrap_or_else(|| "clang++".to_string()),
```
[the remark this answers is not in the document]
tronical: 👍 Will do

skia-bindings/build_support/skia/config.rs
tronical: Hmm, I'm not sure that's possible because `split_once()` returns a tuple and we're interested in only the first field of the tuple. Or is there a way to resolve against just the function even though it doesn't accept a tuple?

skia-bindings/build_support/skia/config.rs
tronical: Uhh, good point! It could be the last parameter. I'll fix that.

skia-bindings/build_support/skia_bindgen.rs
tronical: Good point, that would be much cleaner. I'll change that.

skia-bindings/build_support/skia_bindgen.rs
tronical: Oops yes, I'll point fingers to rust-analyser ;)

tronical:
> * It would be great if there is a separate section in the README.md in skia-bindings that describes the environment variables that are expected to be set if a build with the Yocto SDK is intended (copying the text from the PRs description would be fine for a first draft). Perhaps a short description how to set up the SDK would be useful, too.

ACK.

> * If perfectionism calls, some of the build parameters, like sysroot may be pulled into `FinalBuildConfiguration`, too.

I'll take a look at that.

> * Code comments would be helpful where environment variables override others.

Agreed.

> * The most important point is to ensure that this does not break in the future. I am willing to maintain the changes, but can do so only with help and a GitHub workflow that builds a common / Slints feature configuration against the Yocto SDK (gl + textlayout?). This may be a separate PR, and I am willing to create a suitable workflow if you point me to instructions on how to set up a Yocto SDK build environment.

The part about building against a Yocto SDK inside GH actions is hard, I think. I've been wanting something like that for a while, but oddly I couldn't find anyone doing something similar. Maaaaybe it means manually building an SDK (because a build from sources would take way to long for GH actions), uploading it somewhere and then the workflow here would just download the SDK (mine here is ~2GB 7z, but it contains a lot of junk), extract it (takes a few minutes) and then build.

I don't know yet when I'll get to that, but for tracking purposes I've also created https://github.com/slint-ui/slint/issues/1456 on our side.

tronical:
I think I've got everything covered now, apart from the CI bit.

skia-bindings/build_support/skia/config.rs
pragmatrix: I meant to just remove the `crate::build_support::` qualification prefix, because the module `cargo` is already in scope.

tronical: Got it, thanks :)

pragmatrix:
Thank you @tronical, this looks good now. I'm merging this and prepare a new release.

PR #679, opened 2022-08-15
Comment converter

pragmatrix:
This PR adds a executable project named `comment-converter` that helps converting Skia C++ comments to Rust comments. To convert selections directly in VSCode, the extension [FilterText](https://marketplace.visualstudio.com/items?itemName=yhirose.FilterText) can be used.

The comment-converter is not perfect, but does ~90% of the manual conversion that is needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,8 @@ members = [
     "skia-bindings",
     "skia-safe",
     "skia-org",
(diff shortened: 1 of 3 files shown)


ctrlcctrlv:
Good news! I'm glad you got around to doing more than thinking about it as I did :laughing:

PR #684, opened 2022-08-31
Add support for precompiled Windows binaries that have D3D, GL and textlayout enabled

tronical:
I know that there are infinite combinations of features :-)

For Slint we use Skia's text layout feature (naturally) and we default to using D3D. We have a use-case where somebody wants to develop an app that embeds some OpenGL code and combine that with the Skia renderer (for good text rendering quality), on Windows and Linux. They are okay with depending on OpenGL drivers on Windows and the Skia GL backend works well already. But the Rust Skia bindings are of course complicated by the fact that for this combination no binaries are available, and especially on Windows the Skia build is ... complicated :-). Hence this attempt at a compromise.

I would've filed this as a "wish list" issue first, but I realise that with your mk-workflows helper this is actually quite easy to do, so I just made a patch. But feel free to close this PR if you disagree with this or if this makes the binaries overall too big.

In case this is okay, then the only part I'm uncertain about is if the order (d3d,gl) is correct or if it should be the other way around, to correctly match what the download will attempt when it constructs the file name/url.

### DIFF
--- a/.github/workflows/windows-release.yaml
+++ b/.github/workflows/windows-release.yaml
@@ -929,3 +929,105 @@ jobs:
         artifactErrorsFailBuild: true
         token: ${{ secrets.RUST_SKIA_RELEASE_TOKEN }}
         prerelease: true
+  windows-release-release-d3d-gl-textlayout:
+    runs-on: windows-2022
+    # Containers are not supported on Windows.
+    # container: ghcr.io/pragmatrix/rust-skia-windows:latest
+    env: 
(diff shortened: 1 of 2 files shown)


pragmatrix:
Perfect, thank you a lot, and merging. Will prepare a release soon.

When a string is used to define a `Feature` set, the order is not relevant. It subsequently gets ordered alphabetically in mk-workflows, and similarly for the feature tags in the binaries files.

PR #727, opened 2022-11-09
fix(android): disable outline atomics when using ndk 23+ arm64

triniwiz:
When building with the newer ndk (23+) running on android with an arm64 it fails with the following `cannot locate symbol "__aarch64_ldadd4_acq_rel"` error, seems like it is due to them enabling the flag by [default ](https://reviews.llvm.org/D93585).

I've found similar issues irc they reverted to an older sdk or disabled it

[Flutter](https://github.com/flutter/flutter/issues/75348)
[Chromium](https://chromium.googlesource.com/chromium/src.git/+/refs/heads/main/build/config/compiler/BUILD.gn#1252) (they went a step further)


pragmatrix:
Thank you. I took the liberty to make two cosmetic changes in your branch, merging now.

PR #735, opened 2022-11-22
Support "all features" Linux binaries

### LINKED ISSUE #730: Support to all features in a prebuilt
Hi
Due to the bad network environment (I'm from Cuba), it is impossible for me to compile the library for the features I want to test, I would like to know if it is possible to include a precompiled file that includes all the features.
I am currently trying to use `gl,textlayout,vulkan,x11,egl,wayland`.
I also think it would be good for people who first just want to test the library.
Thanks

pragmatrix:
Closes #730


pragmatrix:
windows-x86 builds fail because of a problem in rstest or rust nightly: https://github.com/la10736/rstest/issues/169

PR #744, opened 2022-12-13
Fix compilation with meta-rust in Yocto

tronical:
When compiling in Yocto with the upstream meta-rust instead of meta-rust-bin, the Yocto target specs are used for compilation. That means the target strings aren't necessarily triplets (or longer) in the form of x86_64-unknown-linux-gnu, but they can be also short like x86_64-linux.

Therefore adjust parse_target() to accomodate that format.

### DIFF
--- a/skia-bindings/build_support/cargo.rs
+++ b/skia-bindings/build_support/cargo.rs
@@ -137,19 +137,29 @@ pub fn host() -> Target {
 pub fn parse_target(target_str: impl AsRef<str>) -> Target {
     let target_str = target_str.as_ref();
     let target: Vec<String> = target_str.split('-').map(|s| s.into()).collect();
-    assert!(target.len() >= 3, "Failed to parse TARGET {}", target_str);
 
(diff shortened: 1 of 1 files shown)


pragmatrix:
Merging, and if you want these shorter / longer targets to keep working in future updates, I'd suggest to add one or two test cases in an additional PR.

tronical:
> Merging, and if you want these shorter / longer targets to keep working in future updates, I'd suggest to add one or two test cases in an additional PR.

Thanks for the quick merge. With test do you mean a bunch of unit tests with `#[test]`?

I have a follow-up change for this one coming up as well, as way past the successful build process I noticed a library linkage problem.

pragmatrix:
> With test do you mean a bunch of unit tests with #[test]?

Yes, indeed, one that tests your usecase is sufficient. I don't want to break anything in future refactorings.

tronical:
This turned out to be uglier than I thought, but it... kind of works: https://github.com/rust-skia/rust-skia/pull/746

PR #745, opened 2022-12-14
Fix external library linkage (freetype, etc.) when building against Yocto

tronical:
A typical target spec when building against Yocto is the poky distribution, which is the vendor (so "aarch64-poky-linux"). When building with meta-rust, instead of meta-rust-bin, that's used instead of the previous aarch64-unknown-linux-gnu style. This results in the target determination in platform.rs to think it's a generic system, which means that we don't end up linking fontconfig and freetype. Those however are a "must". Therefore soften up the match for Linux to allow for any vendor, instead of just "unknown".


tronical:
If preferred I can also change this to explicitly match `poky`, that's a common distro name. But I think ignoring the vendor is better with these standard libraries.

pragmatrix:
Let's go with the more generic variant first, we can refine this later when issues come up.

PR #746, opened 2022-12-15
Enable unit tests in build_support and add target parsing tests

tronical:
The duplicate and mod path hackery is somewhat unfortunate :(


tronical:
Closed in favor of #767

PR #760, opened 2023-01-18
Update Glutin & Winit to version 0.30 where glutin does not include w…

freqmod:
Update the opengl example and its dependencies to use modern version of glutin.

Some of the code around the expected_frame_length_seconds are inspired by neovide.

A lot of the updated code is inspired by the opengl example in glutin (v0.3).

The code is only tested on Wayland in Linux, so some work may be needed to make it work on all the platforms it works on in master.


pragmatrix:
Well done @freqmod  

I've tested in on Windows and macOS and removed iOS support because glutin 0.30 does not support iOS.

PR #762, opened 2023-01-24
Add gl::Interface::new_load_with_cstr

tronical:
This is a kind of overload to new_load_with where the "get_proc_address" function accepts a &CStr instead of a &str. Latest glutin/glow now support get_proc_address with a &CStr, so together with this function it's possible to avoid the round-trip cstring -> str -> cstring.


pragmatrix:
Yes, I like that, thank you! Merging.

PR #767, opened 2023-02-09
Enable testing for build_support and add target parsing tests

pragmatrix:
Followup to #746 from @tronical which adds and enables tests for skia-binding's build_support. This PR fixes the clippy problems and runs the tests as a separate test project.


tronical:
Thanks!! Oops, somehow I didn't see the clippy warnings. I'll close the PR.

Issue #782, opened 2023-03-12
Rendering glitch


marc2332:
hey! It's me again

I am [updating Freya](https://github.com/marc2332/freya/pull/108) to the v0.60.0, so far so good! Except, I found a problem:

![image](https://user-images.githubusercontent.com/38158676/224555653-4ea8c935-df3b-4a5c-8b89-f82c917a585b.png)
( NOW | BEFORE)

As you can see, the only difference is a slim line in the bottom.

I thought it might have been my fault at first, but then I went and tried with skia-safe manually, and it also happened

![image](https://user-images.githubusercontent.com/38158676/224555750-d38e7a8e-6395-457a-800e-f10788f7b1c7.png)

Maybe I am doing something wrong, I don't know, maybe it's a bug on Skia itself?. Here is the code from above: https://github.com/marc2332/rust-skia-rendering-glitch

Thank you again! 😄

pragmatrix:
Thank you a lot for the report and creating the MRE repo. I am not seeing a line on Windows (2019 MBP) and neither on macOS M2. 🤔 Also checked the Skia bugtracker and found nothing recent that looks related.

marc2332:
> Thank you a lot for the report and creating the MRE repo. I am not seeing a line on Windows (2019 MBP) and neither on macOS M2. 🤔 Also checked the Skia bugtracker and found nothing recent that looks related.

Alright, I have made more testing

https://github.com/marc2332/rust-skia-rendering-glitch/tree/old-glutin

![image](https://user-images.githubusercontent.com/38158676/224575586-1e3bf610-c90d-4142-9ccb-cd21e3e1a0b9.png)


I just tried running skia-safe v0.60.0 with glutin v0.28.0 and... it works! I also tried v0.58.0 with glutin v0.30.0, and it also fails.


Doesn't matter if it's 0.58.0 or 0.60.0, seems like an issue with glutin 0.30.0 🤔 

Perhaps it's a bug on glutin v0.30.0? Or maybe i am doing something wrong?

marc2332:
I have also just tried my repo, Skia 0.60.0 with glutin 0.30.0 on Linux and it works great.

So far, looks like a glutin v0.30.0 + Windows problem, athough you tried and it worked fine 🤔 What windows version do you have? I am using Windows 11

pragmatrix:
> What windows version do you have?

A Windows 10 22H2 with all updates installed.

marc2332:
[...]

Issue #764, opened 2023-01-27
`Surface::from_backend_render_target` returns `None`


Speykious:
I was trying to create a Skia hello-world window using glutin's latest API, which dissociates glutin from winit and raw-window-handle, but I couldn't arrive at a working window.
Instead, it panics with the following:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', examples/mao-blur.rs:53:6
```

## Dependencies

```toml
[dependencies]
glow = "0.12.0"
glutin = "0.30.3"
glutin-winit = "0.2.2"
raw-window-handle = "0.5.0"
skia-safe = { version = "0.58.0", features = ["x11", "gl"] }
winit = "0.27.5"
```

## Code

```rs
use std::error::Error;
use std::ffi::CString;
use std::num::NonZeroU32;

use glow::HasContext;
use glutin::config::{Config, ConfigTemplateBuilder};
use glutin::context::{ContextApi, ContextAttributesBuilder, Version};
use glutin::display::GetGlDisplay;
use glutin::prelude::{GlConfig, GlDisplay, NotCurrentGlContextSurfaceAccessor};
use glutin::surface::{GlSurface, SurfaceAttributesBuilder};
use glutin_winit::ApiPrefence;
use raw_window_handle::HasRawWindowHandle;
use skia_safe::gpu::gl::FramebufferInfo;
use skia_safe::gpu::{BackendRenderTarget, SurfaceOrigin};
use skia_safe::{Color, ColorType};
use winit::event::{Event, WindowEvent};
use winit::window::{Window, WindowBuilder};

// Guarantee the drop order inside the FnMut closure. `Window` _must_ be dropped after
// `DirectContext`.
//
// https://github.com/rust-skia/rust-skia/issues/476
struct SkiaEnv {
    surface: skia_safe::Surface,
    gr_context: skia_safe::gpu::DirectContext,
    window: Window,
}

fn create_surface(
    window: &Window,
    gl_config: &Config,
    fb_info: FramebufferInfo,
    gr_context: &mut skia_safe::gpu::DirectContext,
) -> skia_safe::Surface {
    let size = window.inner_size();
    let backend_render_target = BackendRenderTarget::new_gl(
        (
            size.width.try_into().unwrap(),
            size.height.try_into().unwrap(),
        ),
        gl_config.num_samples() as usize,
        gl_config.stencil_size() as usize,
        fb_info,
    );
    skia_safe::Surface::from_backend_render_target(
[...]

PR #783, opened 2023-03-14
gl-window example: Use the minimum number of samples

pragmatrix:
See #764 and #782. Since Skia takes care of anti-aliasing, MSAA is usually not recommended. This PR changes the configuration selection logic in the gl-window example to prefer an available configuration with the minimum number of samples.


tronical (approved):
I did exactly the same fix in other projects.

PR #785, opened 2023-03-20
fix example: Only resize when size is > 0

marc2332:
Hey Armin

What do you think about my changes? 

`NonZeroU32` panics when the value is 0, and this happens when you minimize the window, with my changes it works now :)

### DIFF
--- a/skia-safe/examples/gl-window/main.rs
+++ b/skia-safe/examples/gl-window/main.rs
@@ -255,7 +255,7 @@ fn main() {
                     );
                     /* First resize the opengl drawable */
                     let (width, height): (u32, u32) = physical_size.into();
-                    
+
                     env.gl_surface.resize(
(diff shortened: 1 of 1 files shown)


pragmatrix:
Thank you Marc. With this change, the surface keeps the size before the minimization. Should't it be resized to max(width, 1) and max(height, 1), so that the associated memory is freed?

marc2332:
> Thank you Marc. With this change, the surface keeps the size before the minimization. Should't it be resized to max(width, 1) and max(height, 1), so that the associated memory is freed?

That sounds better 😅Sorry 😁 

just pushed!

marc2332:
Seems like the CI randomly failed 😵‍💫🥴😁😆

pragmatrix:
Looks like networking problems.

PR #791, opened 2023-03-28
feat: add Codec::get_frame_info() and export Codec::NoFrame

Eywek:
Hello, 

I've started to add missing methods from `SkCodec` to be able to use it for decoding GIF formats. 

I've exported `SkCodec_kNoFrame` and updated `Options.prior_frame` to be able to use it when using `get_image` on a GIF. 

I've also tried to export `get_frame_info` method (skia reference [here](https://skia.googlesource.com/skia/+/refs/heads/chrome/m112/src/codec/SkCodec.cpp#793)) but generated bindings export the wrong return type:

```rust
extern "C" {
    #[link_name = "\u{1}__ZN7SkCodec12getFrameInfoEv"]
    pub fn SkCodec_getFrameInfo(this: *mut SkCodec) -> u8;
}
```

I haven't really used `binding` so if you know how to resolve this, I'm listening

Thank you

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,6 +16,7 @@
         "emsdk",
         "fembed",
         "Fira",
+        "frameinfo",
         "fvisibility",
         "heightmaps",
         "isystem",
--- a/skia-bindings/build_support/skia_bindgen.rs
+++ b/skia-bindings/build_support/skia_bindgen.rs
@@ -633,6 +633,9 @@ const ENUM_TABLE: &[EnumEntry] = &[
     ("ColorSpace", rewrite::k_xxx),
     // m109: SkGradientShader::Interpolation::HueMethod
     ("HueMethod", rewrite::k_xxx),
+    // SkCodecAnimation
+    ("DisposalMethod", rewrite::k_xxx),
+    ("Blend", rewrite::k_xxx),
 ];
 
 pub(crate) mod rewrite {
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -7,6 +7,7 @@
 // codec/
 #include "include/codec/SkEncodedOrigin.h"
 #include "include/codec/SkCodec.h"
+#include "include/codec/SkCodecAnimation.h"
 // core/
 #include "include/core/SkAnnotation.h"
 #include "include/core/SkBlendMode.h"
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -166,6 +166,10 @@ extern "C" int C_SkCodec_getFrameCount(SkCodec* self) {
     return self->getFrameCount();
 }
 
+extern "C" void C_SkFrameInfo_Construct(SkCodec::FrameInfo* uninitialized) {
+    new (uninitialized) SkCodec::FrameInfo();
+}
+
 extern "C" bool C_SkCodec_getFrameInfo(SkCodec* self, int index, SkCodec::FrameInfo* info) {
     return self->getFrameInfo(index, info);
 }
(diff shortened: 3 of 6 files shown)


Eywek:
> Usually a wrapper prefixed with `C_` is used to interface with Skia, see [`master`/skia-bindings/src/bindings.cpp#L105-L171](https://github.com/rust-skia/rust-skia/blob/master/skia-bindings/src/bindings.cpp?rgh-link-date=2023-03-28T15%3A01%3A25Z#L105-L171) for the already existing SkCodec wrapper C functions.

Right, thank you!

So I've added the `C_` interface with the right return type (do I need to create a another struct for FrameInfo or is it okay to expose the Sk one?)

I'm stuck now with the following code:
```rust
let infos = SkCodec_FrameInfo {};
unsafe { sb::C_SkCodec_getFrameInfo(self.native_mut(), index, &mut infos) };
```

because I need to init a `SkCodec_FrameInfo` and pass it as ref to the C++ method for populating; but obviously the Rust compiler wants me to initialize each member of `SkCodec_FrameInfo`, is here any way to avoid bothering with defaults here?

Thank you

pragmatrix:
> do I need to create a another struct for FrameInfo or is it okay to expose the Sk one

For FrameInfo it looks that it can be represented as a simple copyable type, which I prefer to mirror on the Rust side to be able to control visibility and member naming (there are exceptions where you could use the bindgen generated ones, but these require field name rewriting in the binding generator). These types are then `NativeTransmutable`, for example  the SkMatrix wrapper: https://github.com/rust-skia/rust-skia/blob/53d08abc6b664bf9a83c8f8fd4c81d2131d93360/skia-safe/src/core/matrix.rs#L33-L40. `

>because I need to init a SkCodec_FrameInfo and pass it as ref to the C++ method for populating; but obviously the Rust compiler wants me to initialize each member of SkCodec_FrameInfo, is here any way to avoid bothering with defaults here?

There is a way by using the `construct` function in Rust and on the C++ side a placement new. For Copy, aka C++ struct types, it's probably safe to leave them uninitialized, but I usually use a default and pass the mutable pointer to it with `.native_mut()`.

Here is an example with placement new (every native transmutable type implements `construct`):

Rust:

https://github.com/rust-skia/rust-skia/blob/53d08abc6b664bf9a83c8f8fd4c81d2131d93360/skia-safe/src/core/m44.rs#L729-L732

C++ 

https://github.com/rust-skia/rust-skia/blob/53d08abc6b664bf9a83c8f8fd4c81d2131d93360/skia-bindings/src/bindings.cpp#L931-L933

And here one with a default:

https://github.com/rust-skia/rust-skia/blob/53d08abc6b664bf9a83c8f8fd4c81d2131d93360/skia-safe/src/codec/_codec.rs#L67-L71

Eywek:
I guess I need to implement a `native_frameinfo` method to convert `SkCodec::FrameInfo` to `FrameInfo`?

pragmatrix (changes requested):

pragmatrix:
I took the liberty to add the remaining types and fields needed for `FrameInfo`, which should make the layout testcases pass, and also changed the type `prior_frame` to `Option<usize>`, `None` meaning `NoFrame`.

Eywek:
> I took the liberty to add the remaining types and fields needed for `FrameInfo`, which should make the layout testcases pass, and also changed the type `prior_frame` to `Option<usize>`, `None` meaning `NoFrame`.

Great! Thank you, lgtm

Eywek:
Hello,

Do you know when you'll be able to release this? 

Thank you

pragmatrix:
Thank you for the ping, binaries are on its way, after that I'll release the Milestone 113 update as soon as possible.

PR #796, opened 2023-04-10
Support building not from a git repo

antonsmetanin:
Added support for building and running examples when the repo is not cloned using git but downloaded as a ZIP archive instead.

Currently the build script assumes that if we're not building as a crate, we must be in a git repo and tries to run the `git submodules update` command. If that fails, the build script panics.
The merge request checks the exit code of the git command and, in case of a failure, falls back to downloading the dependencies as TAR archives.

Another problem was, when the repo is downloaded as a ZIP, the directories for dependencies (`skia` and `depot_tools`) exist but are empty, so I added a check to make sure they also contain something.

### DIFF
--- a/skia-bindings/build_support/binary_cache/download.rs
+++ b/skia-bindings/build_support/binary_cache/download.rs
@@ -100,7 +100,7 @@ fn download_dependencies() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


pragmatrix:
Merging! Thank you a lot @antonsmetanin

PR #807, opened 2023-05-25
Fix memory corruption when creating YUVABackendTextures

tronical:
The C++ constructor expects an array of GrBackendTexture. Since #751 the BackendTexture type in Rust is not a GrBackendTexture anymore though, but a refhandle, so we were passing a pointer.

Work around this by not calling the C++ constructor directly anymore but use a helper function in gpu.cpp that convers the array of pointers to an array of GrBackendTexture objects.


pragmatrix:
Argh, should have checked all usages in #751. Thanks a lot.

PR #811, opened 2023-06-15
fix: Default `thickness_multiplier` to 1.0 in `Decoration`

marc2332:
Having a `0.0` (which is what Rust puts by default) in `thickness_multiplier` will crash the process. So, better to leave it as `1.0` which probably makes more sense, right?

image

### DIFF
--- a/skia-safe/src/modules/paragraph/text_style.rs
+++ b/skia-safe/src/modules/paragraph/text_style.rs
@@ -69,7 +69,7 @@ impl Default for Decoration {
             mode: TextDecoration::default(),
             color: Color::default(),
             style: TextDecorationStyle::default(),
-            thickness_multiplier: 1.0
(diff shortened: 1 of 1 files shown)


pragmatrix:
In #810 I've marked `decoration_mut` as deprecated to align with Skia's `TextStyle.h` file, and `TextStyle` sets `thickness_multiplier` to 1.0. So right, the default should surely be 1.0. But now that `Decoration` can't be set _at once_ anymore, I wonder if we should add back a convenience function `set_decoration_all` (`set_decoration` is taken, it sets only `TextDecoration`, alas the "type" only), or make the `get_` and `set_` functions symmetric and diverge further from the C++ API.

Moreover `Decoration` should actually be `Option<Decoration>` or even a more complex enum on the Rust side.

PR #830, opened 2023-08-30
Upgrade llvm on Windows

pragmatrix:
Latest nightly builds seem to fail because llvm is pre-installed on Windows images and choco install does not upgrade to LLVM 16, which - as it seems - Skia needs to build its C++ libraries.


tronical:
Ahh yes, you're also affected. Lots of projects are it seems. Everyone is waiting for https://github.com/actions/runner-images/issues/8125 :)

.github/workflows/supplemental-builds.yaml
```diff
@@ -107,8 +107,8 @@ jobs:
     - name: Python Version
       run: python --version
 
-    - name: Install LLVM
-      run: choco install llvm
+    - name: Upgrade LLVM
+      run: choco upgrade llvm
```
tronical, line 111: This works around the exception, but yields new C++ error later then.

```suggestion
      run: |
          choco upgrade llvm
          # Skia's highest_version_dir.py script can't handle a directory that's not fully versioned
          Move-Item -Path "C:\Program Files\LLVM\lib\clang\16" -Destination "C:\Program Files\LLVM\lib\clang\16.0.6"
```

pragmatrix:
@tronical Thank you _a lot_ for the links, research, and code suggestions, but as long there is no immediate demand for a release, I am sitting this out until the new Windows image is available and then take a closer look.

.github/workflows/supplemental-builds.yaml
pragmatrix: I've took another approach for the next release. I've removed the need for Skia to call the python script to locate the version  number by setting `clang_win_version` gn argument from within build.rs, this way there is no need for further maintenance when the directory changes again. And luckily there is also no need  for `/D_CRT_USE_BUILTIN_OFFSETOF` anymore in the latest Skia releases.

I am aware that you might need to get older rust-skia Versions running using Slint, so this might be more difficult there. I, for my part, try to get out a release as soon as possible.

tronical: That’s awesome, a new release would solve our problems.

tronical: >  And luckily there is also no need for /D_CRT_USE_BUILTIN_OFFSETOF anymore in the latest Skia releases.

I still see that offsetof issues in my local builds. Do you know where in Skia this need was removed?

pragmatrix: No, I just assumed that, because the problem never appeared in the local and the CI builds.

tronical: Ahh, that's good to know. Thanks for the quick response :).

PR #848, opened 2023-10-12
Fix cross-compiling with system freetype libraries

tronical:
Override the system freetype include path to make sure it's picked up from the sysroot (via = prefix).

### DIFF
--- a/skia-bindings/build_support/platform.rs
+++ b/skia-bindings/build_support/platform.rs
@@ -33,10 +33,17 @@ pub fn link_libraries(features: &Features, target: &Target) -> Vec<String> {
     details(target).link_libraries(features)
 }
 
+pub fn filter_features(target: &Target, features: Features) -> Features {
+    details(target).filter_platform_features(features)
+}
+
 pub trait PlatformDetails {
     fn gn_args(&self, config: &BuildConfiguration, builder: &mut GnArgsBuilder);
     fn bindgen_args(&self, _target: &Target, _builder: &mut BindgenArgsBuilder) {}
     fn link_libraries(&self, features: &Features) -> Vec<String>;
+    fn filter_platform_features(&self, features: Features) -> Features {
+        features
+    }
 }
 
 #[allow(clippy::type_complexity)]
--- a/skia-bindings/build_support/platform.rs
+++ b/skia-bindings/build_support/platform.rs
@@ -33,15 +33,23 @@ pub fn link_libraries(features: &Features, target: &Target) -> Vec<String> {
     details(target).link_libraries(features)
 }
 
-pub fn filter_features(target: &Target, features: Features) -> Features {
-    details(target).filter_platform_features(features)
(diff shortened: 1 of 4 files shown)


tronical:
On a related note, I have a setup now of two Yocto SDKs that can be used in a CI (500MB downloads each). I'd be happy to add a CI build here as well to try to cover this cross-compilation stuff with all the `CLANGXX` variables, etc. that we added a while ago. The one thing I'm unsure about is where to place those 500MB files. Would a repo with git-lfs make sense?

pragmatrix:
Regarding the Yocto SDK, I have to take a closer look what it actually does, but it seems specifically used for cross compiling to embedded platforms? If so, I am not sure if this should be part of the binaries provided by rust-skia. With all the feature combinations, it sounds like this would introduce a lot more complexity to the CI and release process and the benefit is doubtful because I suspect that deployments to embedded platforms need specific adjustments that depend on the project's requirements and are not able to be fulfilled by generic binaries. And furthermore, the binaries provided by rust-skia are for improving the development experience, which - I assume - happens primarily on desktop platforms even when targeting embedded ones. 

If your experience differs, let me know. And thank you a lot for the offer.

Regarding this PR, the beta checks should be fixed by rebasing it on master.

tronical:
> Regarding the Yocto SDK, I have to take a closer look what it actually does, but it seems specifically used for cross compiling to embedded platforms? If so, I am not sure if this should be part of the binaries provided by rust-skia. With all the feature combinations, it sounds like this would introduce a lot more complexity to the CI and release process and the benefit is doubtful because I suspect that deployments to embedded platforms need specific adjustments that depend on the project's requirements and are not able to be fulfilled by generic binaries. And furthermore, the binaries provided by rust-skia are for improving the development experience, which - I assume - happens primarily on desktop platforms even when targeting embedded ones.

Oh yes, I completely agree. Providing binaries for embedded platforms doesn't really make sense given the diversity of the ecosystem. I was bringing up the Yocto SDK <-> CI connection solely from a QA perspective: A way to detect any possible build issues related to cross compilation to embedded platforms before a new release, such as `--sysroot` related things. I remember had "promised" that long time ago [here](https://github.com/rust-skia/rust-skia/pull/668#issuecomment-1204859418), so I wanted to follow up :)

> 
> If your experience differs, let me know. And thank you a lot for the offer.
> 
> Regarding this PR, the beta checks should be fixed by rebasing it on master.

Cool, I'll rebase the PR.

tronical:
The emscripten build error suggests that previously the free type headers from the host were used in the build. Now whatever is in the emscripten sysroot is found first and that seems to be different than what Skia expects. I wonder if perhaps use_system_freetype should not be used with the emscripten build?

pragmatrix:
This looks ready to me. I am merging this so that it will be released alongside the Milestone 119 update.

tronical:
Thank you!

PR #852, opened 2023-10-13
Fix SKIA_USE_SYSTEM_LIBRARIES on Linux

hack3ric:
SKIA_USE_SYSTEM_LIBRARIES doesn't work as expected in two ways:

- When building with system libraries, dependencies like wuffs still needs to be pulled:

  ```
  ninja: error: '../../../../../../../../../.cargo/git/checkouts/rust-skia-archrv-ea99b69400d8833f/7ad5220/skia-bindings/skia/third_party/externals/wuffs/release/c/wuffs-v0.3.c', needed by 'obj/third_party/externals/wuffs/release/c/libwuffs.wuffs-v0.3.o', missing and no known rule to make it
  ```

- If `SKIA_USE_SYSTEM_LIBRARIES=true` and `FORCE_SKIA_BUILD=true`, `binary_cache::resolve_dependencies` will not be called, leaving the intended Skia source directory empty when calling `gn gen`:
  
  ```
  thread 'main' panicked at /build/.cargo/registry/src/index.crates.io-6f17d22bba15001f/skia-bindings-0.62.0/build_support/skia/config.rs:315:10:
    gn error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  ```

- System dependencies isn't actually linked, causing errors like:

  ```
  /usr/bin/ld: /usr/lib/libpng16.so.16: error adding symbols: DSO missing from command line
  ```

  Adding all possible dependencies to `link_libraries` would fix, and should not link to unused libraries (`-Wl,--as-needed` is on by default)

### DIFF
--- a/skia-bindings/build_support/platform/linux.rs
+++ b/skia-bindings/build_support/platform/linux.rs
@@ -34,5 +34,26 @@ pub fn link_libraries(features: &Features) -> Vec<String> {
         }
     }
 
+    if skia::env::use_system_libraries() {
+        libs.push("png16");
+        libs.push("z");
+        libs.push("icudata");
+        libs.push("icui18n");
+        libs.push("icuio");
+        libs.push("icutest");
+        libs.push("icutu");
+        libs.push("icuuc");
+        libs.push("harfbuzz");
+        libs.push("expat");
+
+        if features.webp_encode || features.webp_decode {
+            libs.push("webp");
+        }
+    }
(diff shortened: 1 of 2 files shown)


pragmatrix:
All the changes look reasonable. Thank you for clarifying the behavior of SKIA_USE_SYSTEM_LIBRARIES. Merging now.

PR #873, opened 2023-11-10
Don't remove padding from base64 embeddings in SVG

### LINKED ISSUE #872: Raster embedded in an SVG fails to render
Here is the code I use to render an SVG:
```rust
canvas.clear(Color::WHITE);
let file = File::open("./embedded_raster.svg").unwrap();
let reader = BufReader::new(file);
let svg = SvgDom::read(reader).unwrap();
svg.render(canvas);
```
This SVG file doesn't get rendered:

romanzes:
This fixes #872 by removing the checks and sanitisation that are performed in `decode_base64()`. I believe they aren't needed because the same checks are done by the `base64` crate, and, in case of padding sanitisation, it's actually harmful since `base64` treats strings with removed padding as invalid.

I added tests for the edge cases that are targeted by the removed logic.

### DIFF
--- a/skia-safe/src/modules/svg.rs
+++ b/skia-safe/src/modules/svg.rs
@@ -208,28 +208,8 @@ fn decode_base64(value: &str) -> Vec<u8> {
         .chars()
         .filter(|&c| !is_html_space(c))
         .collect::<String>();
-    let mut input = &*without_spaces;
 
-    if input.len() % 4 == 0 {
-        if input.ends_with("==") {
-            input = &input[..input.len() - 2]
-        } else if input.ends_with('=') {
-            input = &input[..input.len() - 1]
-        }
-    }
-
-    if input.len() % 4 == 1 {
-        return Vec::new();
-    }
-
-    if input
-        .chars()
-        .any(|c| c != '+' && c != '/' && !c.is_alphanumeric())
-    {
-        return Vec::new();
-    }
-
-    match base64::decode(input) {
+    match base64::decode(&without_spaces) {
         Ok(bytes) => bytes,
         Err(_) => Vec::new(),
     }
--- a/skia-safe/src/modules/svg.rs
+++ b/skia-safe/src/modules/svg.rs
@@ -234,8 +234,8 @@ mod base64 {
 
 #[cfg(test)]
 mod tests {
-    use crate::Canvas;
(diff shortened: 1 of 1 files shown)


skia-safe/src/modules/svg.rs
```diff
@@ -208,28 +208,8 @@ fn decode_base64(value: &str) -> Vec<u8> {
...
-    if input.len() % 4 == 1 {
-        return Vec::new();
-    }
-
-    if input
-        .chars()
-        .any(|c| c != '+' && c != '/' && !c.is_alphanumeric())
-    {
-        return Vec::new();
-    }
```
romanzes, line 230: The same thing is checked by the `base64` crate [here](https://github.com/marshallpierce/rust-base64/blob/7f81bafe0e76bcdb84f3069d81d56d77a7a24280/src/engine/general_purpose/decode.rs#L77-L79).

skia-safe/src/modules/svg.rs
```diff
@@ -208,28 +208,8 @@ fn decode_base64(value: &str) -> Vec<u8> {
...
         .collect::<String>();
-    let mut input = &*without_spaces;
 
-    if input.len() % 4 == 0 {
-        if input.ends_with("==") {
-            input = &input[..input.len() - 2]
-        } else if input.ends_with('=') {
-            input = &input[..input.len() - 1]
-        }
-    }
```
romanzes, line 219: Due to the removal of padding here, `base64` crate fails to parse the string because the config `rust-skia` creates has `DecodePaddingMode::RequireCanonical` by default: https://github.com/marshallpierce/rust-base64/blob/7f81bafe0e76bcdb84f3069d81d56d77a7a24280/src/engine/general_purpose/mod.rs#L262

romanzes: An alternative solution would be to create the config with explicit `DecodePaddingMode::Indifferent`, but I don't see why padding has to be removed before passing the string to the `base64` crate.

pragmatrix:
@romanzes Looks good to me, appreciate the test cases. Can you please run `cargo fmt`, I can't write to your branch for some reason. Thanks.

romanzes:
> @romanzes Looks good to me, appreciate the test cases. Can you please run `cargo fmt`, I can't write to your branch for some reason. Thanks.

Done!

Issue #814, opened 2023-06-18
Multiple `&mut` references to the same `Canvas` are possible


ennis:
Version: 0.62.0, haven't checked the latest master yet

It's possible to have multiple exclusive references to the same `Canvas` at the same time, by cloning the handle to the owning surface:

```
let mut surf : Surface = Surface::new(...);  // placeholder, returns Surface=RCHandle<SkSurface>
let mut surf2 = surf.clone();
let canvas1 = surf.canvas();
let canvas2 = surf2.canvas();
eprintln!("canvas1={:p}, canvas2={:p}", canvas1, canvas2);
```

The code above prints `canvas1=0x23d0bf29080, canvas2=0x23d0bf29080`. Having two exclusive refs to the same object in that way is unsound. There might be more instances of this pattern in places where RCHandle is used, but I haven't looked in detail yet.

I think that the `canvas` method should borrow the `SkSurface` directly and not the RCHandle. However, this would require introducing interior mutability with RefCell on the Surface type(`type Surface=RCHandle<RefCell<SkSurface>>`), which has consequences on the ergonomy of the API.

pragmatrix:
I've thought about this for a while and today I was also looking at the WebGPU APIs in Rust, there I've recognized that most handle types (queue, device, surface, etc) can be used through shared references. This gave me the idea that interior mutability can be treated as residing in the C++ side opaque to the Rust compiler.

So this means that instead of returning `&mut Canvas` from a Surface, returning `&Canvas` and rewriting all functions from expecting `&mut self` to `&self` may be a viable strategy here. It will most likely break existing code (though causing only marginal porting effort), but also would make a number of use-cases feasible that Rust prevented before. Canvas (and Surface also) are also neither Send & Sync, so I guess this should be safe and sound.

PR #876, opened 2023-11-14
Canvas::draw_shadow() should receive a shared self reference.

syrel:
Canvas::draw_shadow() should receive a shared self reference.
See [rust-skia/rust-skia#814] and [rust-skia/rust-skia#816]


pragmatrix:
Thank you @syrel. I've missed to update references in impl blocks that are outside the canvas module, and there are more... . Also the clippy problems with beta releases will be fixed separately. Merging.

PR #901, opened 2023-12-14
switch from ureq to curl for downloading binaries

coderedart:
succeeding #900 
discussion at #897


pragmatrix:
Merging, I will remove the commented code as soon the next milestone release gets adopted with positive feedback about the curl based downloads.

Thank you for the contribution!

PR #907, opened 2023-12-19
Typo in Typeface::from_data deprecation warning

warrd:
The function that should be used is called `FontMgr::new_from_data` not `FontMgr::make_from_data`


pragmatrix:
Yes, that's correct, thank you, merging.

PR #915, opened 2023-12-28
Review d3d example

pragmatrix:
This is a follow up to #894, focusing on improving the D3D example:

- Closing the window ends the application.
- Support ESC to close the Window.
- Some more uses / imports to get the code easier on the eyes.
- Remove OnceCell, Mutex, Arc.
- Actually move the rectangle up when the arrow up key was pressed.

/cc @namse


namse:
LGTM

PR #918, opened 2024-01-02
Added some missing comments to `wrap_backend_render_target`.

FledgeXu:
I noticed that `wrap_backend_render_target` is missing some comments about arguments and it would be nice to have them. :-D


pragmatrix:
Thank you @FledgeXu! Merging.

PR #919, opened 2024-01-05
Enable `framebufferOnly` for `CAMetalLayer`.

FledgeXu:
Enabled `framebufferOnly` for `CAMetalLayer` in example `metal-window` to support blend mode.
More discussions see:
https://groups.google.com/g/skia-discuss/c/FTO0NSR_09Q
https://developer.apple.com/documentation/quartzcore/cametallayer/1478168-framebufferonly

### DIFF
--- a/skia-safe/examples/metal-window/main.rs
+++ b/skia-safe/examples/metal-window/main.rs
@@ -51,6 +51,7 @@ fn main() {
         layer.set_device(&device);
         layer.set_pixel_format(MTLPixelFormat::BGRA8Unorm);
         layer.set_presents_with_transaction(false);
+        layer.set_framebuffer_only(false);
 
         unsafe {
             let view = match raw_window_handle {


FledgeXu:
I haven't tested other backends, maybe they also have the same problem?

pragmatrix:
Thank you. Please add a comment to that line that explains when and why this flag should be set to false. This isn't actually relevant for the example itself, but important for users that build their code based on that.

> I haven't tested other backends, maybe they also have the same problem?

Probably, a good way to test this would be to unify the examples and render something that would clearly show that the blending works.

FledgeXu:
> Thank you. Please add a comment to that line that explains when and why this flag should be set to false. This isn't actually relevant for the example itself, but important for users that build their code based on that.

I just updated it.

> Probably, a good way to test this would be to unify the examples and render something that would clearly show that the blending works.

Ok, Let's merge this first.

PR #936, opened 2024-02-15
Fix filter property propagation

romanzes:


romanzes:
Sorry, wrong repo again

PR #954, opened 2024-03-19
Simplify d3d-window example.

Greatness7:
Removes around 80 lines of what I believe to have been unnecessary code, much of which was unseemly pointer arithmetic.

Also fixes what may potentially have been an error in the example: The `surfaces::wrap_backend_render_target` function specifies that the caller must ensure the `backend_render_target` is valid for the lifetime of the `Surface`, but the example was just immediately dropping it after calling said function. Though I may be misunderstand the requirement there.


pragmatrix:
The new d3d example looks really compact and clean now. Thank you @Greatness7. Merging...

PR #1001, opened 2024-07-08
Attempt to fix Linux CI jobs

pragmatrix:
This PR attempts to fix the following CI problem (Happening because of https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/):

```
/__e/node20/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)
```

- https://github.com/actions/checkout/issues/1809

What seems to work is to use an alternative checkout action that does not use node.js:

- https://github.com/actions/checkout/issues/1809#issuecomment-2210575518

### DIFF
--- a/.github/workflows/linux-qa.yaml
+++ b/.github/workflows/linux-qa.yaml
@@ -23,7 +23,7 @@ jobs:
       SKIA_DEBUG: 0
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: taiki-e/checkout-action@v1
       with:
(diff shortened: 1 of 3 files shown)


pragmatrix:
Closing, Linux images were updated to Ubuntu 20.04, see https://github.com/pragmatrix/rust-skia-containers/pull/11

PR #1002, opened 2024-07-13
Fix build with coretext

tronical:
As per release notes, SkShaper::MakeCoreText() is now SkShapers::CT::CoreText().


tronical:
Maybe I should clarify. The compilation works, but the resulting binary doesn't link because of undefined reference to the old function name. This only comes up at application link time though, since it's just an archive.

pragmatrix:
Thank you Simon.

PR #1023, opened 2024-09-01
Add Vizia release job binaries

pragmatrix:
This adds some more vulkan release binaries, as requested in #961 @Greatness7 .

@marc2332


Greatness7:
Thanks👍

Issue #1014, opened 2024-08-07
Improve SVG support


Aiving:
Currently, `skia-safe` provides relatively little of the SVG functionality that Skia has.

There is nothing wrong with this, but the current implementation is only suitable for fairly simple SVG rendering use cases. When, for example, the `fill="currentColor"` attribute starts to be used, there is no way to set a color value for `currentColor` other than the default. As a result, `currentColor` becomes just an alias for black.

Also the following functions are not available in `Dom` struct: `get_root`, `render_node`, `container_size` (although `set_container_size` is there), `find_node_by_id`. As a consequence, there are also no structs like `SvgNode`, `Svg` and `PresentationContext`. Actually adding them is not a complicated process. *Almost*. I've been trying to implement it myself for a couple of days already. The main headache in this case is the `SkSVGProperty` class in Skia, which for some reason in bindings turns into `type SkSVGProperty_ValueT<T> = T` and is not even used anywhere.

I would also like to note that I still don't quite understand how `render_node` and `find_node_by_id` should be used, as they both take as one of the parameters the `id` of a node, the value of which for, for example, root node, I don't know.

DerCommander323:
I too hope that SVG support can be improved, as I currently can't find a way to resize an SVG to an absolute size because it's not possible to get the actual dimensions of the SVG.

geom3trik:
> I too hope that SVG support can be improved, as I currently can't find a way to resize an SVG to an absolute size because it's not possible to get the actual dimensions of the SVG.

You can do `svg.inner().fContainerSize.fWidth` and `svg.inner().fContainerSize.fHeight` to get its size if specified.

DerCommander323:
> > I too hope that SVG support can be improved, as I currently can't find a way to resize an SVG to an absolute size because it's not possible to get the actual dimensions of the SVG.
> 
> You can do `svg.inner().fContainerSize.fWidth` and `svg.inner().fContainerSize.fHeight` to get its size if specified.

Oh, thank you so much. I was looking at Skia's api docs which say to use `getRoot()->intrinsicSize()` which I thought was missing from the bindings.

HinTak:
render_node is used by ot-svg fonts. You'll need to get one of those...

See the skia-python example:
https://github.com/rougier/freetype-py/blob/master/examples/ot-svg-example-skia.py

Which loads this skia-python svg code:
https://github.com/rougier/freetype-py/blob/master/examples/skia_ot_svg_module.py

[...]

Issue #1003, opened 2024-07-17
Make external image loading optional


HookedBehemoth:
I tried a library which used skia-safe and I was surprised to find that it pulls in ureq for a seemingly niche feature in the svg loader. Do you think you could make this part optional even when building with the svg feature?
I believe such an external asset loading should be handed back to the caller to decide what to do with it. With the current implementation authorization on the requested assets isn't possible either.

I'd be happy to provide a patch myself if you could tell me how you want it to be implemented.

https://github.com/rust-skia/rust-skia/blob/aa8c37c3e718c3df4ad28678f244d02490484095/skia-safe/src/modules/svg.rs#L122-L148

Greatness7:
I've noticed that `ureq` and the dependencies it brings (like `rustls` and `idna`) are rather expensive in terms of binary size and are routinely showing up in the top 10 of `cargo bloat` on my applications. As well as causing significant overhead in compile times. 

Considering I only intend to use static resources bundled with the program, including an HTTP client and TLS library and all the related baggage seems very unnecessary.

I'd also like to see these dependencies made optional if possible. @pragmatrix

pragmatrix:
Beginning with version 0.80 of skia-safe, `ureq` is now a separate feature that can be enabled optionally. For the simplest usage of using the SVG Dom, only a (default or empty) FontMgr needs to be provided. Details are in #1024.

Issue #906, opened 2023-12-19
Add an ability to provide custom functions for resource loading


dmtrKovalenko:
My use case: I have already available decoded images as pure binaries that I can load directly from memory instead of using the filesystem. Now this functions are hardcoded while it would be very helpful if I will be able to implement them myself

https://github.com/rust-skia/rust-skia/blob/03ef1ef138b5870c26b8c38d8c356d90cbecde39/skia-safe/src/modules/svg.rs#L171

pragmatrix:
Beginning with version 0.80 of skia-safe, it is possible to implement and use custom resource providers directly in Rust. More details are in #1024. Closing.

Issue #845, opened 2023-10-06
The ureq dependency in the svg feature does not compile wasm


StringKe:
Several of the tls that ureq relies on require the use of ring, a cryptographic processing package, and ring cannot be compiled under wasm at this time.

```
➜ make build                         
EMCC_CFLAGS="-s ERROR_ON_UNDEFINED_SYMBOLS=0 -s MAX_WEBGL_VERSION=2 -s MODULARIZE=1 -s EXPORT_NAME=createRustSkiaModule -s EXPORTED_RUNTIME_METHODS=GL" EMSDK=~/.asdf/installs/emsdk/3.1.46 cargo build --target wasm32-unknown-emscripten
   Compiling ring v0.16.20
   Compiling ureq v2.8.0
error[E0432]: unresolved import `super::sysrand_chunk`
   --> /Users/chen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ring-0.16.20/src/rand.rs:306:16
    |
306 |     use super::sysrand_chunk::chunk;
    |                ^^^^^^^^^^^^^ could not find `sysrand_chunk` in `super`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `ring` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
make: *** [build] Error 101

```

upstream error: https://github.com/briansmith/ring/pull/1568

REPO: https://github.com/StringKe/wasm-error

pragmatrix:
@StringKe You can use the svg in <https://github.com/rust-skia/rust-skia/pull/569#issuecomment-977843719> and [this test case](https://github.com/rust-skia/rust-skia/blob/ef7eebd1ebb9c473d9f92877948ec5eae23387a8/skia-safe/src/modules/svg.rs#L262-L272) shows how to use the API.

pragmatrix:
`ureq` is now an optional dependency and disabled by default. Closing.

PR #1024, opened 2024-09-01
Add bindings for SVG Resource Providers.

pragmatrix:
This PR attempts to properly support SVG related Resource Providers, removes the ureq dependency in the process, and will fix some of the problems with the current wrappers.

This PR is related to the following issues:

- #1014
- #1003 
- #906 
- #845

Finalizing:

- [x] Add a `data:` base64 testcase
  - [x] A malformed one (tested manually)
  - [x] A good one
  - [x] Percent encoding.
- [x] Test on Windows (why on earth is the URL handling different????)
- [x] Is it possible to get rid of the duplicated FontMgr requirement in `Dom`?
- [x] Document how to create a Local / Ureq and `FontMgr` based `NativeResourceProvider`.


pragmatrix:
@triniwiz Initially you implemented the `data:` url handling in the SVG resource providers. Do you have an small example (an image maybe) that works with the Skia Dom, so that I can add them as a full SVG rendering test case?

triniwiz:
@pragmatrix You can use the following [svg](https://upload.wikimedia.org/wikipedia/commons/4/4c/The_Hague%2C_Netherlands%2C_the_old_city_center.svg) for testing.

pragmatrix:
The difference on Windows handling of http(s) urls originates from the the splitting mechanism that is used to separate the path from the filename. The `SkOSPath` functions use system dependent separators for resolving directory and file names:

From `SkSVGImage.cpp`:
```
        case SkSVGIRI::Type::kNonlocal: {
            const auto path = SkOSPath::Dirname(href.iri().c_str());
            const auto name = SkOSPath::Basename(href.iri().c_str());
            imageAsset = rp->loadImageAsset(path.c_str(), name.c_str(), /* id */ name.c_str());
            break;
```

PR #1025, opened 2024-09-04
feat: `svg::Dom::root` + `svg::Dom::DomSVG`

marc2332:
## Changes
- Adds a getter to for the root node of a svg DOM, `svg::Dom::root()`
- The node (`DomSVG`) comes with a `intrinsic_size()` getter and a `set_attribute()` setter.

## Help 👋🏻 
That said, I am struggling to get the RC system to work with the node type `DomSVG`.

```rs
error[E0308]: mismatched types
   --> skia-safe\src\macros.rs:29:27
    |
29  |             let _: &$nt = &(a._base);
    |                    ----   ^^^^^^^^^^ expected `&SkRefCnt`, found `&SkSVGContainer`
    |                    |
    |                    expected due to this
    |
   ::: skia-safe\src\modules\svg.rs:231:1
    |
231 | require_base_type!(sb::SkSVGSVG, sb::SkRefCnt);
    | ---------------------------------------------- in this macro invocation
    |
    = note: expected reference `&SkRefCnt`
               found reference `&SkSVGContainer`
    = note: this error originates in the macro `require_base_type` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
error: could not compile `skia-safe` (lib) due to 1 previous error
```

Somehow the same thing works perfectly fine with the svg DOM, but not with the svg node, and I can't find the reason 🤔 

Note: I have 0 experience in cpp.


marc2332:
Thanks @pragmatrix ! That fixed both issues. 

Now I must find out why `parseAndSetAttribute` always returns `false`...

marc2332:
I am kinda stuck @pragmatrix, do you have any clue why would `parseAndSetAttribute` always return false? I am a complete noob (or worse than that) when it comes to cpp, so maybe I am just missing something very obvious...

pragmatrix:
I see one obvious problem: Rust `&str` are not null-terminated, they cause undefined behavior when cast to `const char*`. Convert them to a `CString` on the Rust side, before passing them `as_ptr()` like so

https://github.com/rust-skia/rust-skia/blob/5f17a054e21ff0cd5b90146c218a13e3a7f3825d/skia-safe/src/core/font_mgr.rs#L149-L155

marc2332:
Oh, that has to be it... Will try asap :) Still learning, thanks! 😄

marc2332:
This is promising!

marc2332:
I feel like the CI is failing because it is using Rust 1.81

marc2332:
@pragmatrix Can you lend me a hand with the tests when you have some time?

These are failing:

```sh
 failures:
    __bindgen_test_layout_SkTCopyOnFirstWrite_open0_SkSVGLengthContext_close0_instantiation
    __bindgen_test_layout_SkTCopyOnFirstWrite_open0_SkSVGPresentationContext_close0_instantiation
    __bindgen_test_layout_SkTLazy_open0_SkPath_close0_instantiation
    __bindgen_test_layout_SkTLazy_open0_SkSVGViewBoxType_close0_instantiation
    bindgen_test_layout_SkSVGRenderContext
    bindgen_test_layout_SkSVGSVG
```

pragmatrix:
Two types were causing defective layouts, I made them opaque. For more clippy beta errors, rebase on master.

skia-safe/examples/svg/main.rs
marc2332, line 1: I suppose you probably don't want this example right? 😄 I used it to test my changes

pragmatrix: If this can be reformulated as an actual test (for example if the intrinsic size changes when attributes are changed) You can add this as a test to svg.rs if you want.

skia-safe/src/modules/svg.rs
```diff
@@ -221,6 +221,100 @@ impl Dom {
...
+    CM,
+    MM,
+    IN,
+    PT,
+    PC,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub struct DomSVGLength {
```
pragmatrix, line 308: Rename to SvgLength to keep this aligned with the Skia naming? Or even, for purity, just `Length`. This is open for discussion. I usually start with the purest names (we are already in the module `svg`) and then see how client code will look like. Sometimes I then get back to adding _redudant_ context to the names.

skia-safe/src/modules/svg.rs
```diff
@@ -221,6 +221,100 @@ impl Dom {
...
+    pub fn set_height(&mut self, height: DomSVGLength) {
+        unsafe {
+            sb::C_SkSVGSVG_setHeight(self.native_mut(), *height.native());
+        }
+    }
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum SvgUnit {
```
pragmatrix, line 292: nit: I would rather see this directly taken from the bindings and just `use SkSVGLength_Unit as Unit` an add a rule (see  skia_bindgen.rs ENUM_REWRITES) to strip the `k` prefix from the variant names. This way we don't have to adjust the Rust code if members change.

pragmatrix (changes requested):

Aiving:
I've also decided to contribute to SVG support, so I'd like to show the [current result of a week of work](https://github.com/marc2332/rust-skia/compare/feat/svg-dom-utils...Aiving:rust-skia:feat/svg-dom-utils).

pragmatrix:
@Aiving This is awesome. From a first look this includes everything Marc has done and a lot more. Is that what exists now working stable? @marc2332 Can you take a look at @Aiving branch and see if this will meet your needs, and if so - and assuming @Aiving wants to contribute his changes here - I suggest to close this one and open a new PR.

Aiving:
> @Aiving This is awesome. From a first look this includes everything Marc has done and a lot more. Is that what exists now working stable? @marc2332 Can you take a look at @Aiving branch and see if this will meet your needs, and if so - and assuming @Aiving wants to contribute his changes here - I suggest to close this one and open a new PR. 
> 

Regarding stability: many layout tests fail (will probably have to add some types to opaque). On top of that I'm pretty sure it won't compile at the moment, since I changed the function signature from C++ at the last moment and didn't change the code in Rust in any way. However, I will fix it as soon as I get the chance!

Aiving:
However, as for the code's workability - it does work!

image

image

image

marc2332:
@pragmatrix  I can happily push @Aiving commits in my branch, then I will apply the changes you requested

Sounds good to both of you?

pragmatrix:
@marc2332 actually no. The reason is that there are a number of essential differences in the approach of your solutions, and @Aiving's looks much more sophisticated and complete, but also introduces a number of different tradeoffs (introducing a separate macro crate, etc.), which _will cause_ discussions and issues that are not related to this PR, and I am not willing to get this bundled up (code wise or contributor wise). 

I'm fine with merging this PR first and then attending to @Aiving's solution when he feels ready for a PR. But as it looks right now to me (correct me me if I am wrong!), there won't be much left of this one.

marc2332:
> @marc2332 actually no. The reason is that there are a number of essential differences in the approach of your solutions, and @Aiving's looks much more sophisticated and complete, but also introduces a number of different tradeoffs (introducing a separate macro crate, etc.), which _will cause_ discussions and issues that are not related to this PR, and I am not willing to get this bundled up (code wise or contributor wise).
> 
> I'm fine with merging this PR first and then attending to @Aiving's solution when he feels ready for a PR. But as it looks right now to me (correct me me if I am wrong!), there won't be much left of this one.

Well, that's sad to hear that because I would have hoped to finally make some useful contribution to the project. But I also understand that if @Aiving changes are better, then this PR is not really needed.

Closing this

PR #1034, opened 2024-09-24
support SkRuntimeShaderBuilder, SkFilters::RuntimeShader

nongio:
This PR adds a new method to the image_filters: `runtime_shader` to expose the functionality of `SkImageFilters::RuntimeShader`
as well `SkRuntimeShaderBuilder` is now wrapped.
RuntimeShaderBuilder have 2 helper methods to set all the supported uniform types (float and int)

## Usage:
```rust
let sksl_code = r#"
  uniform shader content;
  uniform float4 uColor;
  
  half4 main(float2 coords) {
      half4 cColor = content.eval(coords);
      return half4(mix(cColor, uColor, 0.5));
  }
"#;

// Compile the SkSL shader
let mut runtime_effect = RuntimeEffect::make_for_shader(sksl_code, None).unwrap();

// Create a shader builder
let mut builder = RuntimeShaderBuilder::new(&mut runtime_effect);

// Set the uniform color
let color = [0.0, 1.0, 0.0, 1.0]; // Green color
builder.set_uniform_float("uColor", &color);

let filter_shader = image_filters::runtime_shader(&builder, "", None);
```

### DIFF
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -2682,7 +2682,6 @@ void C_SkRuntimeShaderBuilder_setUniformFloat(SkRuntimeShaderBuilder *self,
   using float2 = std::array<float, 2>;
   using float3 = std::array<float, 3>;
   using float4 = std::array<float, 4>;
-  using float2x2 = std::array<float, 4>;
   using float3x3 = std::array<float, 9>;
   using float4x4 = std::array<float, 16>;
 
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -2644,6 +2644,93 @@ bool C_SkRuntimeEffect_allowBlender(const SkRuntimeEffect* self) {
     return self->allowBlender();
 }
 
+
+void C_SkRuntimeShaderBuilder_Construct(SkRuntimeShaderBuilder *uninitialized,
+                                        SkRuntimeEffect *effect) {
+  new (uninitialized) SkRuntimeShaderBuilder(sp(effect));
+}
+const SkRuntimeEffect *
+C_SkRuntimeShaderBuilder_effect(const SkRuntimeShaderBuilder *self) {
+  return self->effect();
+}
+sk_sp<const SkData>
+C_SkRuntimeShaderBuilder_uniforms(const SkRuntimeShaderBuilder *self) {
+  return self->uniforms();
+}
+SkSpan<const SkRuntimeEffect::ChildPtr>
+C_SkRuntimeShaderBuilder_children(const SkRuntimeShaderBuilder *self) {
+  return self->children();
+}
+SkShader *
+C_SkRuntimeShaderBuilder_makeShader(const SkRuntimeShaderBuilder *self,
+                                    const SkMatrix *localMatrix) {
+  auto shader = self->makeShader(localMatrix);
+  return shader.release();
+}
+
+void C_SkRuntimeShaderBuilder_setUniformFloat(SkRuntimeShaderBuilder *self,
+                                              const char *name, size_t count,
+                                              const float *const values,
+                                              size_t len) {
+  using float2 = std::array<float, 2>;
+  using float3 = std::array<float, 3>;
+  using float4 = std::array<float, 4>;
+  using float2x2 = std::array<float, 4>;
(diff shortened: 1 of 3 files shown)


skia-safe/src/effects/runtime_effect.rs
```diff
@@ -406,3 +406,58 @@ impl ChildPtr {
...
+                name.len(),
+                data.as_ptr() as _,
+                data.len(),
+            )
+        };
+    }
+    pub fn set_uniform_int(&mut self, name: impl AsRef<str>, data: &[i32]) {
+        let name = name.as_ref();
+        unsafe {
+            sb::C_SkRuntimeShaderBuilder_setUniformFloat(
```
tronical, line 454: I suppose this should call the Int variant?:)

nongio: :man_facepalming: indeed thanks for checking, fixed

nongio:
@tronical i can take care of the linting problems later today, on a different note, if you have time to test it, i have a doubt on the:
```
impl NativeDrop for sb::SkRuntimeShaderBuilder
```
doing a simple test where i replace a used builder with a new one is generating a segfault, i assume there a still references to the old builder somewhere in skia, do you know what is the right strategy to apply here?

tronical:
Not sure, off-hand. This could also be a bug in Skia. The docs say that this is experimental and convenience. Perhaps it would make more sense to expose SkRuntimeEffect or this Rust API directly top of it?

nongio:
I did not find another way of implementing it using SkRuntimeEffect, there is some logic about the uniforms attached to the builder and therefore it is required at the moment

tronical:
I thought the builder would basically use this public API like this:

1. Create the RuntimeEffect using https://api.skia.org/classSkRuntimeEffect.html#a9d91ca7d9ce6b64d66d71862f67ec710
2. Make the shader from it with uniforms using https://api.skia.org/classSkRuntimeEffect.html#a1baa99ee870f0a5150ce77ed27b56416
3. Pass the shader to the remaining Skia pipeline.

nongio:
i will double check later, from what i remember the problem is that an image filter will have 1 uniform about the incoming image that is automatically populated, i suppose this logic is in the builder, if not i will try a different approach

nongio:
@pragmatrix is there anything i can do to fix the windows builds?

pragmatrix:
@nongio It seems that the function `SkRuntimeShaderBuilder_SkRuntimeShaderBuilder_destructor` wasn't generated on Windows by bindgen. I guess it's best to wrap the destructor in C_ functions in bindings.cpp, too.

nongio:
alright thanks, i added a method in bindings.cpp

nongio:
this PR should be good for merging, is there anything else that you think i should do?

skia-bindings/src/bindings.cpp
```diff
@@ -2644,6 +2644,99 @@ bool C_SkRuntimeEffect_allowBlender(const SkRuntimeEffect* self) {
...
+}
+
+void C_SkRuntimeShaderBuilder_setUniformInt(SkRuntimeShaderBuilder *self,
+                                            const char *name, size_t count,
+                                            const int *const values,
+                                            size_t len) {
+  using int2 = std::array<int, 2>;
+  using int3 = std::array<int, 3>;
+  using int4 = std::array<int, 4>;
+  switch (len) {
```
tronical, line 2722: For this and the other method, if the provided slice of ints/floats in Rust doesn't match the expected length, the function will just quietly do nothing. That may make it difficult to debug.

I suggest to document the supported slice sizes in the Rust API docs of `set_uniform_float` and `set_uniform_int`, and perhaps even return an error if no size matched.

nongio: makes sense, will include the suggestions

nongio: @tronical  the PR is updated, the rust method is documented and now return a Result

tronical: Love it, the custom error version looks great

pragmatrix:
This looks good to me. But before merging, I suggest to remove the unused wrappers `C_SkRuntimeShaderBuilder_effect`, `C_SkRuntimeShaderBuilder_uniforms` and `C_SkRuntimeShaderBuilder_children`.

PR #1066, opened 2024-12-03
Vulkan window example

samizdatco:
This PR adds a new `vulkan-window` example to `skia-safe/examples`, complementing the existing windowing examples for OpenGL, Metal, and D3D. It's admittedly pretty bulky by comparison, but perhaps unavoidably so given how many low-level pieces Vulkan requires you to configure before getting started.

The approach taken here is a mixture of the Vulkano [triangle](https://github.com/vulkano-rs/vulkano/blob/master/examples/triangle/main.rs) demo, @pragmatrix's [Emergent](https://github.com/pragmatrix/emergent/blob/master/src/renderer.rs) renderer, and the excellent [Skulpin](https://github.com/aclysma/skulpin/tree/master/skulpin-renderer/src) crate (which the maintainer has sadly moved on from at this point). Most of the comments from the triangle demo have been left in place to explain the voluminous setup boilerplate, but could be stripped out or condensed to make the code less of a wall-of-text.

One thing that I'm too much of a Vulkan neophyte to figure out is how to combine what is (I think?) currently a 2-pass rendering process into one. At the moment, the DirectContext does its rasterization, hands the resulting bitmap back to the cpu, then sends it back to the gpu through the swapchain. But I get the impression that Skulpin's texture shader-based approach was able to have skia's output go directly to the framebuffer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,30 @@ Please share your build experience so that we can try to automate the build and 
 
 ## Example Applications
 
+### gl-window
+
+An example that opens an OpenGL Window and draws the rust-skia icon with skia-safe (contributed by [@nornagon](https://github.com/nornagon)).
+
+```bash
+cargo run --example gl-window --features gl
+```
+
+On Linux the feature `x11` needs to be enabled:
+
+```bash
+cargo run --example gl-window --features gl,x11
+```
+
+### vulkan-window
+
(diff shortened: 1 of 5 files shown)


pragmatrix:
> At the moment, the DirectContext does its rasterization, hands the resulting bitmap back to the cpu, then sends it back to the gpu through the swapchain.

Going over the code, I don't see any indication that the CPU is involved in rendering (as far as I understand, the Framebuffer, and Surface types are just wrapper types and all target the underlying GPU SwapChain Image). Also comparing the runtime behavior with the metal example seems to indicate, that there is even better GPU utilization if rendering is done every frame. CPU utilization is exactly the same.

I've updated the README.md and will merge this now without further consideration.

Thank you a lot @samizdatco.

Issue #1015, opened 2024-08-09
Alpine: Don't hardcode C++ stdlib version


Janrupf:
https://github.com/rust-skia/rust-skia/blob/5053d4dbf463714df7177b648eacf466eb57d5d4/skia-bindings/build_support/platform/alpine.rs#L27-L32

This causes build failures where the C++ stdlib can't be found on newer alpine installations. Instead the version should be detected automatically.

pragmatrix:
Closed via #1067

PR #1067, opened 2024-12-03
Alpine: use detected rather than hardcoded stdlib version 

samizdatco:
Updated the build script to run `g++ -dumpversion` to determine the proper include paths.

Addresses #1015


pragmatrix:
Looks good, thank you @samizdatco, merging.

Issue #1055, opened 2024-10-31
Support for rendering SVGs with both system fonts and dynamically loaded fonts?


samizdatco:
We've recently been doing some work on Skia Canvas to add support for SVG rendering (samizdatco/skia-canvas#180) and I've been having problems getting font rendering to work correctly. I've been creating Dom objects with the `svg::Dom::from_bytes` method and have had success using system fonts when I pass in a `FontMgr::default()` as the font manager. Likewise, I can use dynamically-loaded fonts by creating a `TypefaceFontProvider`, adding the fonts to it, and passing it as the Dom's FontMgr. What's unclear to me is how to allow the SVG to use both the system fonts **and** the ones I've loaded dynamically.

When working with the Paragraph api, I'm used to creating a `FontCollection` and using that to combine `FontMgr` and `TypefaceFontProvider` instances into a union of all their contents. But I don't see how that's possible with SVGs since only a single FontMgr can be specified. I noticed the `utils::OrderedFontMgr` type and hoped that that might be the solution, but even though it behaves like a normal FontMgr (and even lists all the expected typefaces when queried), passing it to the Dom constructor actually results in **no** fonts rendering correctly. Am I misunderstanding the role of `OrderedFontMgr`, or is this something that actually *should* be working but for an undiscovered bug?

Here's the approach I tried:

```rs
    let ttf_path = "path/to/a/font.ttf";

    // use the default mgr to access system fonts
    let sys_mgr = FontMgr::default();

    // load the local font and add it to a provider
    let mut dynamic = TypefaceFontProvider::new();
    if let Ok(bytes) = fs::read(ttf_path){
        let local_font = sys_mgr.new_from_data(&bytes, None).unwrap();
        dynamic.register_typeface(local_font, None);
    }

    // combine the two into a single mgr
    let mut ord_mgr = skia_safe::utils::OrderedFontMgr::new();
    ord_mgr.append(sys_mgr);
    ord_mgr.append(dynamic);

    // print out all the families and styles, noting that the dynamically loaded one
    // is included at the end of the list, after all the system fonts
    for i in 0..ord_mgr.count_families() {
        let name = ord_mgr.family_name(i);
        println!("font_family: {name}");
        let mut style_set = ord_mgr.new_style_set(i);
        for style_index in 0..style_set.count() {
            let (_, style_name) = style_set.style(style_index);
            if let Some(style_name) = style_name {
                println!("  style: {style_name}");
            }
        }
    }

[...]

PR #1068, opened 2024-12-04
Add temporary replacement for buggy `SkOrderedFontMgr` class

samizdatco:
The `SkOrderedFontMgr` class has [a bug](https://github.com/rust-skia/rust-skia/issues/1055) in its `onMatchFamily` method where it mistakenly looks for a null response rather than a zero-length `SkFontStyleSet` to indicate "no matches" from a given `FontMgr`. As a result it never looks in any of the FontMgrs it contains aside from the first one (totally defeating the purpose of the class which is supposed to act as a ‘union’ of multiple FontMgrs).

 I've reported [the issue](https://issues.skia.org/issues/382016481) upstream but it would be *extremely* useful to me to have a working OrderedFontMgr in skia-safe in the meantime. The actual fix is just a two line change to a single method, but unfortunately the original class locks its list of FontMgrs behind a private reference so I wasn't able to figure out a way to replace the broken method via subclassing.

Instead I've had to inllne the full implementation as `RustOrderedFontMgr` in `bindings.cpp`. I left the existing SkOrderedFontMgr accessors in place so this temporary class can be removed if and when Skia itself is patched. At that point, the references to `RustOrderedFontMgr` in `ordered_font_mgr.rs` could also be turned back into `SkOrderedFontMgr`.

I've also added a unit test to `ordered_font_mgr.rs` that verifies fonts can be found in FontMgrs that aren't the first in the list. It relies on the `textlayout` feature to create single-font FontMgrs via `TypefaceFontProvider`, but if that's a problem it may be possible to rewrite it with fewer dependencies…

### DIFF
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -3172,6 +3172,140 @@ extern "C" void C_SkParsePath_ToSVGString(const SkPath* self, SkString* uninitia
 }
 
 //
+// Temporary replacement for buggy SkOrderedFontMgr
+//
+
+class RustOrderedFontMgr : public SkFontMgr {
+public:
+    RustOrderedFontMgr();
+    ~RustOrderedFontMgr() override;
+
+    void append(sk_sp<SkFontMgr>);
+protected:
+    int onCountFamilies() const override;
+    void onGetFamilyName(int index, SkString* familyName) const override;
+    sk_sp<SkFontStyleSet> onCreateStyleSet(int index)const override;
+    sk_sp<SkFontStyleSet> onMatchFamily(const char familyName[]) const override;
+    sk_sp<SkTypeface> onMatchFamilyStyle(const char familyName[],
+                                         const SkFontStyle&) const override;
+    sk_sp<SkTypeface> onMatchFamilyStyleCharacter(const char familyName[], const SkFontStyle&,
+                                                  const char* bcp47[], int bcp47Count,
+                                                  SkUnichar character) const override;
+
+    sk_sp<SkTypeface> onMakeFromData(sk_sp<SkData>, int ttcIndex) const override;
+    sk_sp<SkTypeface> onMakeFromStreamIndex(std::unique_ptr<SkStreamAsset>,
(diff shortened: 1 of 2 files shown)


samizdatco:
One additional change: I implemented the `onLegacyMakeTypeface` method, which was (intentionally?) left as a no-op in the upstream version of the class. With this in place, it's now possible to use an OrderedFontMgr as the FontMgr for an SVG ResourceProvider, giving it access to fonts from multiple sources (e.g., the system font library plus user-loaded font files).

pragmatrix:
Interesting, but I suggest that we cherry-pick your upstream PR https://github.com/google/skia/pull/193 directly into the [rust-skia commits](https://github.com/google/skia/compare/chrome/m131...rust-skia:m131-0.79.1) that are applied on top of every new Skia milestone. The Rust testcases will make sure that it does not get lost.

> I've also added a unit test ... but if that's a problem it may be possible to rewrite it with fewer dependencies…

I think the textlayout and font dependencies are ok for now.

Also, do you need a release with binaries soon? The next scheduled release is Milestone 132 (on January the 8th).

samizdatco:
Sounds great; I'll open a new PR with just the SkOrderedFontManager changes in the [skia](https://github.com/rust-skia/skia) repo.

As for a binary release, I haven't actually been able to use the binaries since adding the `webp` feature. So just adding a tag to the `rust-skia` repo would be enough for me to be able to make use of the changes. If there were any way to add `webp` to future releases though it would cut down on my compile times significantly. 

Here are the features I'm currently using for Skia Canvas on various platforms:
- mac (x86 & arm64): `metal`, `textlayout`, `webp`, `svg`
- linux (x86 & arm64 / glibc & musl): `vulkan`, `embed-freetype`, `freetype-woff2`, `textlayout`, `webp`, `svg`
- windows (x86 only): `vulkan`, `textlayout`, `webp`, `svg`

pragmatrix:
Everything in here is handled by the following PRs #1070, #1082, #1089. Thank you @samizdatco. Closing.

PR #1070, opened 2024-12-05
Incorporate ordered fontmanager fixes by @samizdatco

pragmatrix:
.. directly into our Skia tree, tag is ~~m131-0.80.0~~, m131-0.80.1, see #1068

/cc @samizdatco

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 
 Skia Submodule Status: chrome/m131 ([upstream changes][skia-upstream], [our changes][skia-ours]).
 
-[skia-upstream]: https://github.com/rust-skia/skia/compare/m131-0.79.1...google:chrome/m131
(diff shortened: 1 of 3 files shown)


pragmatrix:
@samizdatco It seems that the two commits do result in a compile error, I had to add the following commit to compensate: [42254e3011a9c3431b732f754763b679310bace6](https://github.com/rust-skia/skia/commit/42254e3011a9c3431b732f754763b679310bace6)

samizdatco:
> @samizdatco It seems that the two commits do result in a compile error, I had to add the following commit to compensate: [42254e3011a9c3431b732f754763b679310bace6](https://github.com/rust-skia/skia/commit/42254e3011a9c3431b732f754763b679310bace6)

Oops, that's left over from when I dropped the arg names due to compiler warnings about unused properties (when the function was a no-op). Thanks for fixing it!

samizdatco:
🎉

Issue #1075, opened 2024-12-16
StructureElementNode child nodes


danbishop-canva:
There seems to be an issue with pointers addresses when retrieving child nodes from a parent node. In `pdf_document.rs`, on line 177, the pointer is assigned with the type `*mut SkPDF_StructureElementNode` but then is passed into `safer::from_raw_parts()` inferred as `*const StructureElementNode`. The retrieved `StructureElementNode` has pointer address of has the original address assigned to `SkPDF_StructureElementNode`. The  `SkPDF_StructureElementNode` contained within `StructureElementNode` now has a different random address. Hopefully the screenshot provides some context.

I tried to figure out a fix but I do not have much experience in FFI. Can anyone see how to figure out a fix for this?

<img width="1728" alt="Screenshot 2024-12-16 at 7 22 49 pm" src="https://github.com/user-attachments/assets/46679e9e-e921-46f9-9331-364f26e83474" />

PR #1076, opened 2024-12-17
Fix retrieving child vector from StructureElementNode

danbishop-canva:
Currently when retrieving child vector, it retrieves the address of the first element in the vector, rather than the address of the pointer to the first element in the vector. This PR fixes the incorrect retrieval logic in the C++ code in the bindings file.

Address #1075

### DIFF
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -3071,18 +3071,26 @@ extern "C" void C_SkPDF_StructElementNode_appendChild(SkPDF::StructureElementNod
     self->fChildVector.push_back(std::unique_ptr<SkPDF::StructureElementNode>(node));
 }
 
-extern "C" size_t C_SkPDF_StructureElementNode_getChildVector(const SkPDF::StructureElementNode *self, SkPDF::StructureElementNode **nodes)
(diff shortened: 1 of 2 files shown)


pragmatrix:
Hmm, I am confused, the implementation of child_vector() in `master` appears to be correct. Initially, I suspected an issue when taking a &* reference on the unique_ptr which may copy the underlying pointer and then returns a reference to a temporary memory location, but a small testcase works fine:

```
    #[test]
    fn structure_element_node_child_vector() {
        let mut root = StructureElementNode::new("root");
        root.append_child(StructureElementNode::new("nested"));
        root.append_child(StructureElementNode::new("nested2"));
        let v = root.child_vector();
        assert_eq!(v[0].type_string(), "nested");
        assert_eq!(v[1].type_string(), "nested2");
    }
```

Is there a minimum reproducible example I can try that causes the problem you've experienced?

BTW: The suggested solution keeps the C++ created array allocated. While this can be solved, I'd rather like to the return a slice to the elements so that the borrow checker makes sure that the node can't be modified as long the array is being accessed.

pragmatrix:
Closing in favor of #1083, which attempts to point directly into the vector and does some static checks to be sure that the types are layout compatible.

PR #1077, opened 2024-12-19
Fix structure element tree

danbishop-canva:
Currently there are `StructureElementNode` is defined in the Rust code but it is not passed through when initialising a PDF. `SetNodeId()` is a function which is used to connect nodes in the Structure Element Tree to elements in the Canvas. The structure element tree is used to generate accessibility tagging in a PDF.


pragmatrix:
Thank you! 👍, merging!

PR #1079, opened 2024-12-20
feat: Support custom ImageAsset

dmtrKovalenko:
Here is an implementation that allows to return your custom implementation of ImageAsset wrapping the rust trait to the skia world using ResourceProvider. ImageAssets are overridable so I added an ability to override it using rust trait object just like you already do for ResourceProvider.

My use case: I have already available decoded images in the memory so I just manually creating the frame_data from the pointer. This is a code:

```rust
struct FFramesSkiaImage(Image);
impl FFramesSkiaImage {
    fn new(image: &ImageData) -> Option<Self> {
        let image_info = ImageInfo::new(
            ISize::new(image.metadata.width as i32, image.metadata.height as i32),
            ColorType::RGBA8888,
            AlphaType::Premul,
            None,
        );

        let sk_image = raster_from_data(
            &image_info,
            unsafe { Data::new_bytes(&image.get_bytes()) },
            image.metadata.width as usize * 4,
        )?;

        Some(Self(sk_image))
    }
}

impl CustomImageAsset for FFramesSkiaImage {
    fn is_multi_frame(&self) -> bool {
        false
    }

    fn get_frame_data(&self, t: f32) -> ImageFrameData {
        create_image_frame_data(
            &self.0,
            Matrix::default(),
            SamplingOptions::default(),
            ImageSizeFit::kFill,
        )
    }
}

impl ResourceProvider for SkiaFFramesProvider {
    fn load_image_asset(
        &self,
        _resource_path: &str,
        resource_name: &str,
        _resource_id: &str,
    ) -> Option<ImageAsset> {
        if self.0.is_null() {
            // TODO log the error
            return None;
        }

        let image = self.0.media_source?.resolve_image(resource_name)?;
        let skia_image = FFramesSkiaImage::new(image)?;

        ImageAsset::from_custom_image_asset(skia_image)
    }

    fn load(&self, _resource_path: &str, _resource_name: &str) -> Option<skia_safe::Data> {
        None
    }

    fn load_typeface(&self, _name: &str, _url: &str) -> Option<skia_safe::Typeface> {
        None
    }

    fn font_mgr(&self) -> FontMgr {
        FontMgr::empty()
    }
}
```

Please let me know if you like the implementation and if you do I'll add the tests covering this API.


skia-safe/src/modules/image_asset.rs
```diff
@@ -0,0 +1,116 @@
...
+        ImageAsset::from_ptr(unsafe {
+            std::mem::transmute::<_, *mut skresources_ImageAsset>(custom_image_asset.into())
+        })
+    }
+}
+
+pub type ImageSizeFit = skresources_ImageAsset_SizeFit;
+pub type ImageFrameData = skresources_ImageAsset_FrameData;
+
+pub fn create_image_frame_data(
```
dmtrKovalenko, line 48: I thought it would be possible to create the FrameData struct as a rust struct but it uses skia shared pointer and it looks like there is no way to get the shared pointer from the `RCHandle` without getting into the skia world. Correct me if I wrong

pragmatrix: Yes, we could provide a generic function for that, but this would in turn mean that we directly handle sp typed smartpointers in Rust, which we don't. There might be a way to create a compatible struct in Rust and placing a RCHandle at the same field position, but this feels scary. Speaking of that, FrameData should be wrapped and made opaque to the Rust world (with functions wrapping access to the fields), otherwise clients could mess around with the smartpointer.

pragmatrix (changes requested):
Thank you a lot for that. I like the addition. 

Here are the things I like to be changed:

- [ ] Put the image_asset.rs into a new subdirectory named modules/resources and refer it from resources.rs (because it is not a module).
- [ ] The `skresources_ImageAsset_SizeFit` enum variants are prefixed with a `k`. Add a `variant_name!` check, and add `SizeFit` to `skia_bindgen.rs` `ENUM_REWRITES`, this should fix it.
- [ ] Wrap the `FrameData` in a `Handle<>`, and offer access to the members through functions (the simple members don't need C wrappers). This is to prevent clients from messing with the sp and also fixes a bug with the current implementation, where the reference count is not reduced when dropped.
- [ ] Add a test case based on a reasonable use case if possible.

dmtrKovalenko:
Thank you for review I have a question though about this 

> Wrap the FrameData in a Handle<>, and offer access to the members through functions (the simple members don't need C wrappers). This is to prevent clients from messing with the sp and also fixes a bug with the cur 

the problem is that it should not be a handle it is stack allocated struct that is used by skia so it needs to be the owned struct when we return it to the skia world. That's why there is no need to drop it or even have access to the fields. If  Correct me if I'm wrong

dmtrKovalenko:
hey @pragmatrix gentle pin on my previous question when you'll had a chance

pragmatrix:
> the problem is that it should not be a handle it is stack allocated struct that is used by skia so it needs to be the owned struct when we return it to the skia world. That's why there is no need to drop it or even have access to the fields. If Correct me if I'm wrong

The `Handle<T>` does not make assumptions about where the value is allocated (compared to `RefHandle` or `RCHandle` which are heap only), and it's `#[repr(transparent)]`, meaning that it exists only as a conceptual abstraction and has no effect on the bits moving around. It's just a way to wrap a C / C++ value so that its contents are hidden from Rust and supports a drop handler that should call back into Skia world and invoke the destructor of that C++ type. 

If it is only passed from Rust to C/C++ (Skia world) it works, but after calling `create_image_frame_data` you are free to drop or modify it, and using `get_frame_data` you can ask for it from the Rust part before converting it to a NativeImageAsset. So as long ImageFrameData gets exported as a public type, this loophole continues to exist and there is no place in skia-safe as far as I know where a `sk_sp<>` is directly exposed.

PR #1081, opened 2024-12-27
chore: Improve working with sendable

dmtrKovalenko:
Currently it is impossible to have Sendabale wrapper on the debuggagble struct + the `unwrap` method is always adds additional anxiety about panicking while by [rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) it should be called `into_inner`

### DIFF
--- a/skia-safe/src/prelude.rs
+++ b/skia-safe/src/prelude.rs
@@ -1,5 +1,6 @@
 #![allow(dead_code)]
 use std::{
+    fmt::Debug,
     hash::{Hash, Hasher},
     marker::PhantomData,
     mem::{self, MaybeUninit},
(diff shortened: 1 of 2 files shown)


pragmatrix:
I think the blanket Clone implementation needs to be removed. E.g. this would allow an RCHandle to be cloned while it continues to be Sendable into another Sendable (having a reference count of 2), and both can be sent and used from different threads then, which would undermine the very nature of why Sendable got introduced in the first place.

dmtrKovalenko:
@pragmatrix aplied the changes

dmtrKovalenko:
this should be ready to review

dmtrKovalenko:
CI issues don't look to be caused by my changes

pragmatrix:
Yes, there is a unrelated problem with wasm builds, I saw this on the nightly scheduled builds. Merging.

PR #1100, opened 2025-02-07
fix the ownership bug to prevent generating a dangling pointer.

### LINKED ISSUE #1099: dangling pointer is created in the pub function legacy_make_typeface() in Skia-safe, leading to a use-after-free bug.
Hi, I've noticed a dangling pointer issue in the pub [`legacy_make_typeface()` function](https://github.com/rust-skia/rust-skia/blob/eb0bf757107e4bd72d325e38940c56d3401223b3/skia-safe/src/core/font_mgr.rs#L233) in skia-safe. 

When the `family_name` parameter is not `None`, it creates a dangling pointer that is passed to the FFI function `sb::C_SkFontMgr_legacyMakeTypeface()`, leading to a Use-After-Free bug.

To reproduce the bug, run the below test code with Valgrind to observe the UAF issue. 
```
use crate::FontMgr;

#[test]
fn create_all_typefaces() {
    let font_mgr = FontMgr::default();
    let families = font_mgr.count_families();
    // This test requires that the default system font manager returns at least one family for now.
    assert!(families > 0);
    let mut style_set = font_mgr.new_style_set(0);
    let family_name = Some("test_family_name");
    let (style, _) = style_set.style(0);
    font_mgr.legacy_make_typeface(family_name, style);
}
```

quitbug:
fix #1099

### DIFF
--- a/skia-safe/src/core/font_mgr.rs
+++ b/skia-safe/src/core/font_mgr.rs
@@ -242,7 +242,7 @@ impl FontMgr {
         Typeface::from_ptr(unsafe {
             sb::C_SkFontMgr_legacyMakeTypeface(
                 self.native(),
-                family_name.map(|n| n.as_ptr()).unwrap_or(ptr::null()),
+                family_name.as_ref().map(|n| n.as_ptr()).unwrap_or(ptr::null()),
                 style.into_native(),
             )
         })


pragmatrix:
Nice find, thank you for the PR!

PR #1115, opened 2025-03-18
Preserve curl outputs

tymokvo:
Hello,

I was trying to use `skia-safe` on my windows machine today and was unable to resolve a build error coming from the skia binaries download. I was able to track this down to the flags used to invoke `curl`.

The `-s` option swallows all output. This means that, even though the build script will report the stderr stream on failure, there is nothing in the string (so I was unable to see the `404`). 

The flag [`--no-progress-meter` was added in `curl 7.67.0`](https://curl.se/docs/manpage.html) to hide the progress stream while keeping any error stream.

The values passed into the `Err` are also still wrapped in `Option<_>`, so I added a bit to unwrap those and make the reported errors a bit cleaner.

### Example output

```console
  --- stdout
  cargo:rerun-if-env-changed=DOCS_RS
  cargo:rerun-if-env-changed=SKIA_DEBUG
  cargo:rerun-if-env-changed=SKIA_SOURCE_DIR
  cargo:rerun-if-env-changed=FORCE_SKIA_BUILD
  cargo:rerun-if-env-changed=FORCE_SKIA_BINARIES_DOWNLOAD
  TRYING TO DOWNLOAD AND INSTALL SKIA BINARIES: 0.83.0/9b7918e9ae25c7529a24-x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=SKIA_BINARIES_URL
    FROM: https://github.com/rust-skia/skia-binaries/releases/download/0.83.0/skia-binaries-9b7918e9ae25c7529a24-x86_64-pc-windows-msvc.tar.gz
  DOWNLOAD AND INSTALL FAILED: curl error code: "22"
  curl stderr: "curl: (22) The requested URL returned error: 404\r\n"
```

### DIFF
--- a/skia-bindings/build_support/binary_cache/utils.rs
+++ b/skia-bindings/build_support/binary_cache/utils.rs
@@ -24,8 +24,8 @@ pub fn download(url: impl AsRef<str>) -> io::Result<Vec<u8>> {
         // fail fast with no "error pages" output. more of a hint though, so we might still get error on stdout.
         // so make sure to check the actual status returned.
         .arg("-f")
-        // silent. no progress or error messages. only pure "response data"
-        .arg("-s")
+        // no progress meter but keep error messages.
+        .arg("--no-progress-meter")
         .arg(url)
         .output();
     match resp {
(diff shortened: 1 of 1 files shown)


pragmatrix:
Thanks a lot @tymokvo

syrel:
For my future self: MacOS Catalina 10.15.7 was shipped with an older curl. In that case the build might fail with the following error:
```
Failed to download https://codeload.github.com/rust-skia/skia/tar.gz/m135-0.83.1 (curl error code: "2"
  curl stderr: "curl: option --no-progress-meter: is unknown\ncurl: try 'curl --help' or 'curl --manual' for more information\n")
```

Developers should update their curl:
```
brew install curl
```

PR #1120, opened 2025-04-10
allow building without pdf support

lunixbochs:


pragmatrix:
Thank you for the PR, but why exactly is it important to you to disable PDF support?

lunixbochs:
the resulting binary ends up about 0.5MB smaller, and this brings skia-safe to parity with the skia build I'm already doing
(I'm in the process of moving the Talon app from SkiaSharp's C bindings to skia-safe)

pragmatrix:
The side effect of this PR is that the binaries will from now on be tagged with "pdf", too. And we won't support non-pdf binaries for the time being. But I think this could be tolerable as long you don't need prebuilt binaries without pdf support. Merging.

skia-safe/Cargo.toml
```diff
@@ -27,7 +27,7 @@ rust-version.workspace = true
 doctest = false
 
 [features]
-default = ["binary-cache", "embed-icudtl"]
+default = ["binary-cache", "embed-icudtl", "pdf"]
```
marc2332, line 30: I am not actually sure about this, why would be a good idea to enable PDF support by default? It is a niche thing IMO

pragmatrix: Fair point. My current thinking is that it would break expectations because we would remove a feature that was previously included in every single binary and I fear that there will be even more requests for binaries with included pdf support. Not sure if it's worth it for half a MB for the usual suspects (Windows, macOS, Linux) that probably stays on disk because of memory mapped code. As a result, removing the default feature would then result in unavailable binaries for most configurations users will chose to experiment with in the beginning. A bit of a conundrum here and I can't see how to separate this, but also don't see it as really important except for heavily space constrained platforms, and on those you usually want to do your own builds anyway.

lunixbochs: doing your own build is feasible enough, I think anyone trying to golf their skia binary size should just do a custom build, and skia-safe should just focus on having good support for custom builds rather than complicating the matrix

marc2332: > Fair point. My current thinking is that it would break expectations because we would remove a feature that was previously included in every single binary and I fear that there will be even more requests for binaries with included pdf support. Not sure if it's worth it for half a MB for the usual suspects (Windows, macOS, Linux) that probably stays on disk because of memory mapped code. As a result, removing the default feature would then result in unavailable binaries for most configurations users will chose to experiment with in the beginning. A bit of a conundrum here and I can't see how to separate this, but also don't see it as really important except for heavily space constrained platforms, and on those you usually want to do your own builds anyway.

Fair point, my pov is just too freya-oriented :)

PR #1121, opened 2025-04-10
add no-compile feature to prevent falling back to a compile

lunixbochs:
I'm packaging and caching correct rust-skia binaries at an earlier stage in my build pipeline - I don't want my app builders to fall back to a full skia build. This feature flag will give me a cargo-level way to ensure my build is working, instead of implicitly rebuilding skia.

### DIFF
--- a/skia-bindings/Cargo.toml
+++ b/skia-bindings/Cargo.toml
@@ -55,6 +55,7 @@ binary-cache = ["flate2", "tar"]
 embed-icudtl = ["lazy_static"]
 embed-freetype = []
 freetype-woff2 = []
+no-compile = []
 
 [dependencies]
 mozjpeg-sys = { version = "2", features = ["with_simd"], optional = true }
--- a/skia-bindings/build.rs
+++ b/skia-bindings/build.rs
@@ -43,6 +43,10 @@ fn main() -> Result<(), io::Error> {
                 None,
             );
         } else {
+            if cfg!(feature = "no-compile") {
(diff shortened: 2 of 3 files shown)


pragmatrix:
I'd rather prefer environment variables to control specific build settings. Would this be reasonable?

lunixbochs:
an extra environment variable doesn't help here, because I specifically want to stop `cargo build` from implicitly starting a skia build

the way skia-safe implicitly builds skia is actually kind of hard to avoid accidentally triggering without a flag like this
let's say my makefile did a release build with the right env vars and then I run a different cargo command without --release, it will start a surprise skia build (with the output suppressed so I can't even tell it's happening besides a long stall and cpu fans)

right now it additionally seems like I can't run any cargo commands without hitting this panic, because my precompile strategy so far was to use the SKIA_BINARIES_URL + FORCE_SKIA_BINARIES_DOWNLOAD and it's hitting
```
cargo:rerun-if-env-changed=FORCE_SKIA_BINARIES_DOWNLOAD
```

certainly it's not great that I can't use cargo commands, but the feature flag is doing the correct thing (as I very much do not want an implicit skia compile to happen in my environment)

lunixbochs:
a better flag might be something like "never try to build skia because we're expecting it to be available in the system search path" (which will work without env vars)

pragmatrix:
Ok, I see your point. Thank you for the detailed explanation. Merging now.

PR #1122, opened 2025-04-11
add feature shared-library

lunixbochs:
supporting a shared library build is useful for multiple reasons

- multiple executables can link against the same skia library in an app
- it's harder to cross compile for windows against a msvc static lib

note I'm still not sure about cross-compiling with a msvc skia-bindings.lib, it would be better if we managed to build that into the main skia dll


lunixbochs:
I think this would work if we got skia-bindings into the skia dll, but as is this won't work and I switched my whole windows build to msvc instead

PR #1124, opened 2025-04-16
feat(openharmony): add openharmony support

richerfu:
This is an experimental support for OpenHarmmony and we can run `hello` example successfully.

Here is example code: https://github.com/richerfu/skia-canvas

image


pragmatrix:
Thank you a lot. Merging.

PR #1127, opened 2025-04-24
Resume interrupted transfer

xb284524239:
When the network condition is poor, the resume-from-breakpoint feature can enhance the user experience.

### DIFF
--- a/skia-bindings/build_support/binary_cache/utils.rs
+++ b/skia-bindings/build_support/binary_cache/utils.rs
@@ -1,5 +1,5 @@
 use std::{
-    fs,
+    fs::{self, File},
     io::{self, Error, ErrorKind, Read},
     path::Path,
 };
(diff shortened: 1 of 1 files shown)


xb284524239:
I don't know why three checks have been cancelled. Do I need to modify the code?

pragmatrix:
The failing checks are unrelated. I'll merge this, but need to add some documentation and also will rename the download function so that it's clear what is does, because it introduces non-trival side-effects now. In addition to the continue-where-leftoff you've basically added a cache, too (by filename that is), which may or may not cause problems I can not foresee. And to remove the possibility for conflicts, I think it's best to put the files in OUT_DIR/.cache/rust-skia. I'll make all these changes after the merge. Thank you a lot!

xb284524239:
Yes, it is necessary to ensure the uniqueness of the file name!

PR #1128, opened 2025-04-27
skia-bindings/build_support: init SKIA_GN_ARGS environment variable

theoheretic:
## What does this PR change?

This PR adds a new environment variable (`SKIA_GN_ARGS`) to be read when building skia-bindings. This should result in a command equivalent to this:
```
gn --args='... ${SKIA_GN_ARGS}' ...
```

## Why?

This could be used when you need to add a new include path using `extra_cflags` without hardcoding it into `build-support/skia/config.rs`.

## Should this break anything?

This shouldn't break anything, or at least I think so.

## Has this been tested?

Yes, and it looks to have worked on my end. The build phase seems to reflect that at least.

```
cargo:rerun-if-env-changed=SKIA_GN_ARGS

Skia args: is_official_build=true is_debug=false skia_enable_svg=false skia_enable_gpu=false skia_enable_skottie=false skia_enable_pdf=true skia_use_gl=false skia_use_egl=false skia_use_x11=false skia_use_system_libpng=false skia_use_libwebp_encode=false skia_use_libwebp_decode=false skia_use_system_zlib=false skia_use_xps=false skia_use_dng_sdk=false skia_use_freetype_woff2=false cc="clang" cxx="clang++" skia_use_icu=false skia_use_harfbuzz=false skia_use_freetype=true skia_system_freetype2_include_path="/does/not/exist" skia_use_system_libjpeg_turbo=false skia_use_expat=true skia_use_system_expat=false target_os="linux" target_cpu="x86_64" extra_cflags=["-I=/usr/include/freetype2","-I/usr/include/freetype2","-O0","--target=x86_64-unknown-linux-gnu"] extra_asmflags=["--target=x86_64-unknown-linux-gnu"] extra_cflags+=["-Ofast"]
```

I have also build Neovide (which depends on skia-safe) and it built fine.

### DIFF
--- a/skia-bindings/build_support/skia/config.rs
+++ b/skia-bindings/build_support/skia/config.rs
@@ -297,12 +297,18 @@ pub fn configure_skia(
     python: &Path,
     gn_command: Option<&Path>,
 ) {
-    let gn_args = build
+    let mut gn_args = build
         .gn_args
         .iter()
         .map(|(name, value)| name.clone() + "=" + value)
         .collect::<Vec<String>>()
         .join(" ");
+
+    if let Some(args) = cargo::env_var("SKIA_GN_ARGS") {
+        gn_args.push_str(
+            format!(" {}", args.as_str()).as_str()
+        );
+    }
 
     let gn_command = gn_command
         .map(|p| p.to_owned())


pragmatrix:
Thank you! I'll add rudimentary documentation to skia-bindings/README.md after the merge.

PR #1139, opened 2025-05-16
add ColorSpace::new_icc()

lunixbochs:
this makes it possible to create a SkColorSpace from an ICC profile, which can be necessary to make a color accurate SkImage from pixel data


pragmatrix:
Because it's not guaranteed that this function will successfully parse the ICC profile, I suggest to return a Option<Self> and let the caller decide how to handle problems instead of panicking right away.

lunixbochs:
turned into Option :+1:

pragmatrix:
Thanks!

Issue #1123, opened 2025-04-14
The vulkan_window example has validation errors


SuchAFuriousDeath:
I'm a contributor of https://github.com/obhq/obliteration.

We use https://github.com/slint-ui/slint, which has an option to use skia with vulkan.

In Obliteration, I get a crash upon closing a popup window. investigating the issue (i.e. turning validation layers on, which leads to the same errors as below) eventually led me here:
https://github.com/slint-ui/slint/blob/master/internal/renderers/skia/vulkan_surface.rs, which is where the rendering with skia is implemented. 

I tried to fix the validation errors, but I cannot for the life of me figure out how. I tried looking for examples of using skia with vulkan, but could't really find anything that would help.

The only thing I found is the example here: https://github.com/rust-skia/rust-skia/tree/master/skia-safe/examples/vulkan-window.

However, when running it with validation layers I get the same validation errors as with Slint.

Since I'm slowly running out of ideas, I'd like to ask here - what is the proper way to ensure image layouts are synchronized correctly with skia? 

I also stumbled upon this: https://github.com/rust-skia/rust-skia/issues/914

The FlushInfo type seems like it would allow passing in a semaphores to be signaled. That is, if the fields weren't private . I suppose one could call the flush function from skia-bindings with the type from that crate,  but I think at least the validation errors in the example deserve an issue.

I apologize in advance if I misunderstood something, I'm not really a GPU person.

The validation errors: 

Validation Error: [ VUID-vkCmdDraw-None-09600 ] | MessageID = 0x46582f7b
vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] command buffer VkCommandBuffer 0x57a6e45b66e0 expects VkImage 0xb000000000b (subresource: aspectMask = VK_IMAGE_ASPECT_COLOR_BIT, mipLevel = 0, arrayLayer = 0) to be in layout VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
[...]

PR #1148, opened 2025-05-29
fix vulkan renderer example not launching and closing incorrectly

bgkillas:
wanted to use this example for my graphing calculator but then it didn't launch so i fixed it, as before it didn't correctly force the color type it wanted to use on creation of the window, and if it wasn't defaulted to it then it crashes, also it wasn't handling closing correctly as it needed to properly drop the vulkan renderer it seems, i also added an example onto how to switch vsync mode to immediate if it is applicable, as the vsync seemed really bad on my calculator for some reason(added alot of latency)

### DIFF
--- a/skia-safe/examples/vulkan-window/main.rs
+++ b/skia-safe/examples/vulkan-window/main.rs
@@ -54,6 +54,7 @@ fn main() {
         ) {
             match event {
                 WindowEvent::CloseRequested => {
+                    self.renderer = None;
                     event_loop.exit();
                 }
                 WindowEvent::Resized(_) => {
--- a/skia-safe/examples/vulkan-window/renderer.rs
+++ b/skia-safe/examples/vulkan-window/renderer.rs
@@ -1,3 +1,4 @@
+//https://raw.githubusercontent.com/rust-skia/rust-skia/refs/heads/master/skia-safe/examples/vulkan-window/renderer.rs
 use ash::vk::Handle;
 use std::{ptr, sync::Arc};
 use vulkano::{
@@ -16,7 +17,9 @@ use skia_safe::{
     gpu::{self, backend_render_targets, direct_contexts, surfaces, vk},
     ColorType,
 };
-
+use vulkano::format::Format;
+use vulkano::image::view::ImageViewCreateInfo;
+use vulkano::image::ImageSubresourceRange;
 use winit::{dpi::LogicalSize, dpi::PhysicalSize, window::Window};
 
 pub struct VulkanRenderer {
@@ -57,8 +60,8 @@ impl VulkanRenderer {
         let (swapchain, _images) = {
             // Querying the capabilities of the surface. When we create the swapchain we can only
             // pass values that are allowed by the capabilities.
-            let surface_capabilities = device
-                .physical_device()
+            let physical_device = device.physical_device();
+            let surface_capabilities = physical_device
                 .surface_capabilities(&surface, Default::default())
                 .unwrap();
 
@@ -66,7 +69,13 @@ impl VulkanRenderer {
             let (image_format, _) = device
                 .physical_device()
                 .surface_formats(&surface, Default::default())
-                .unwrap()[0];
+                .unwrap()
+                .into_iter()
+                .find(|a| a.0 == Format::B8G8R8A8_UNORM)
+                .unwrap();
+            let present_modes = physical_device
+                .surface_present_modes(&surface, Default::default())
+                .unwrap();
 
             // Please take a look at the docs for the meaning of the parameters we didn't mention.
             Swapchain::new(
@@ -105,7 +114,11 @@ impl VulkanRenderer {
                     //
                     // Only `Fifo` is guaranteed to be supported on every device. For the others, you must call
                     // [`surface_present_modes`] to see if they are supported.
-                    present_mode: PresentMode::Fifo,
+                    present_mode: if present_modes.contains(&PresentMode::Immediate) {
+                        PresentMode::Immediate
+                    } else {
+                        PresentMode::Fifo
+                    },
 
                     // The alpha mode indicates how the alpha value of the final image will behave.
                     // For example, you can choose whether the window will be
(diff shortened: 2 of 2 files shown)


bgkillas:
fixes part of #1123

renhiyama:
while this allows the project to start, its too laggy, and even crashes with segfault.
```
Using device: NVIDIA GeForce RTX 4070 (type: DiscreteGpu)
[2025-05-31 15:15:29.348] [MANGOHUD] [info] [gpu.cpp:95] Set renderD128 as active GPU (id=10de:2786 pci_dev=0000:01:00.0)
Failed vulkan call. Error: -4, QueueSubmit(queue, 1, &submitInfo, fence)

thread 'main' panicked at /home/ren/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vulkano-0.35.1/src/sync/future/fence_signal.rs:538:34:
called `Result::unwrap()` on an `Err` value: DeviceLost
[1]    242598 segmentation fault (core dumped)  mangohud cargo run --release

```

Do you know how to fix this?

bgkillas:
wdym by too laggy? i think you are misunderstanding how this project is setup, since its just a simple rectangle, it has no use of redrawing unless its resized, so it wont redraw, if you add renderer.window.request_redraw(); at the end of redraw requested in main i get around 3500 fps using 90% of my igpu, so not that laggy, i dont get this segfault however so i am unable to help with that, might be a vulkano issue from the log but idk

marc2332:
Did you also have this? https://github.com/rust-skia/rust-skia/issues/1151

bgkillas:
unable to replicate @marc2332

skia-safe/examples/vulkan-window/renderer.rs
```diff
@@ -279,8 +293,11 @@ impl VulkanRenderer {
             self.framebuffers = new_images
                 .iter()
                 .map(|image| {
-                    let view = ImageView::new_default(image.clone()).unwrap();
-
+                    let mut info = ImageViewCreateInfo::default();
+                    info.format = Format::B8G8R8A8_UNORM;
+                    info.subresource_range =
+                        ImageSubresourceRange::from_parameters(info.format, 1, 1);
+                    let view = ImageView::new(image.clone(), info).unwrap();
```
samizdatco, line 300: Does this actually have a different result than `ImageView::new_default`? I'm pretty sure it picks up the format from the `image` and the subresource values also  look identical to the defaults (I think?)

skia-safe/examples/vulkan-window/renderer.rs
```diff
@@ -267,6 +280,7 @@ impl VulkanRenderer {
                 .swapchain
                 .recreate(SwapchainCreateInfo {
                     image_extent: window_size.into(),
+                    image_format: Format::B8G8R8A8_UNORM,
```
samizdatco, line 283: This is redundant with the values pulled in on the next line via the existing swapchain's info, right?

skia-safe/examples/vulkan-window/renderer.rs
```diff
@@ -105,7 +114,11 @@ impl VulkanRenderer {
                     //
                     // Only `Fifo` is guaranteed to be supported on every device. For the others, you must call
                     // [`surface_present_modes`] to see if they are supported.
-                    present_mode: PresentMode::Fifo,
+                    present_mode: if present_modes.contains(&PresentMode::Immediate) {
```
samizdatco, line 117: Using immediate mode is definitely 'faster' but will also result in tearing, so it might not be the best thing to default to. Maybe you could add a static boolean in main to let users opt in?

skia-safe/examples/vulkan-window/renderer.rs
```diff
@@ -57,16 +60,22 @@ impl VulkanRenderer {
...
                 .unwrap();
 
             // Choosing the internal format that the images will have.
             let (image_format, _) = device
                 .physical_device()
                 .surface_formats(&surface, Default::default())
-                .unwrap()[0];
+                .unwrap()
+                .into_iter()
+                .find(|a| a.0 == Format::B8G8R8A8_UNORM)
```
samizdatco, line 74: 💯 This definitely needs to match the format requested when creating a surface. One alternative to hard-coding would be to set up [an allowlist](https://github.com/samizdatco/skia-canvas/blob/bbf3094825f85706373b64c9a5ce573e5ccf49e9/src/gpu/vulkan/mod.rs#L9) and then [step through](https://github.com/samizdatco/skia-canvas/blob/bbf3094825f85706373b64c9a5ce573e5ccf49e9/src/gpu/vulkan/renderer.rs#L122) the GPU's supported formats until finding one that matches

skia-safe/examples/vulkan-window/main.rs
```diff
@@ -54,6 +54,7 @@ fn main() {
         ) {
             match event {
                 WindowEvent::CloseRequested => {
+                    self.renderer = None;
```
samizdatco, line 57: 👍

marc2332:
> Did you also have this? #1151

I just realized my issue could because I was using x11, wayland seems to not freeze, what do you folks use?

skia-safe/examples/vulkan-window/renderer.rs
bgkillas: i think this is a fine default to show an example on how to change it given a present mode exists

bgkillas:
> > Did you also have this? #1151
> 
> I just realized my issue could be because I was using x11, wayland seems to not freeze, what do you folks use?
> 
> update: Just so I am clear, this example (even from this PR) also crashes for me under x11 (not under wayland though) when I resize

i use wayland so maybe thats it

skia-safe/examples/vulkan-window/renderer.rs
bgkillas: i dont crash from removing this and the above one so i guess its probably fine, i found the matching issue last so these are just from me being overly cautious i guess

renhiyama:
any update on this guys? (I use wayland btw)

PR #1157, opened 2025-06-16
build: use pkg-config for Linux system libraries

thiblahute:
Replace hardcoded library names with pkg-config detection for better compatibility across Linux distributions and allowing to use libraries that were built and installed in non default directories. We still default to hardcoded library names if pkg-config failed for backward compatibility.


pragmatrix:
I like this, let's see if it goes through the CI with the clippy problem fixed.

pragmatrix:
Thank you @thiblahute. Merging...

PR #1163, opened 2025-06-21
Don't enable FreeType woff2 if FreeType wasn't enabled

pragmatrix:


pragmatrix:
Will be included in #1164. Closing.

PR #1171, opened 2025-07-08
Fix build for archlinux:latest with textlayout feature enabled.

### LINKED ISSUE #1170: Skia-binding fails with textlayout feature
Without the `textlayout` feature I don't have any issues building, but as soon as I add it I get the following error:

```
error[E0412]: cannot find type `std__Rb_tree_color` in this scope
   --> /.../target/debug/build/skia-bindings-30bbc81fb1d2ba99/out/skia/bindings.rs:926:19
    |
926 |     pub _M_color: std__Rb_tree_color,
    |                   ^^^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0412`.
error: could not compile `skia-bindings` (lib) due to 1 previous error
```

This error can be avoided by adding the `all-linux` feature, but any modification to it (except removing gl/egl, since they're a Wayland prerequisite) will cause this error to pop up again. Just `textlayout` on its own with default features also seems to work, but as soon as `egl` is added or default features are removed it seems to break.

pragmatrix:
Closes #1170

### DIFF
--- a/skia-bindings/build_support/skia_bindgen.rs
+++ b/skia-bindings/build_support/skia_bindgen.rs
@@ -481,6 +481,8 @@ const BLOCKLISTED_TYPES: &[&str] = &[
     "SkRuntimeEffect_ConstIterable.*",
     // Linux LLVM9 c++17
     "std::_Rb_tree.*",
+    // archlinux
+    "std::__rb_tree.*",
     // Linux LLVM9 c++17 with SKIA_DEBUG=1
     "std::__cxx.*",
     "std::array.*",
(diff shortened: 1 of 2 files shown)


skia-bindings/build_support/skia_bindgen.rs
```diff
@@ -481,6 +481,8 @@ const BLOCKLISTED_TYPES: &[&str] = &[
     "SkRuntimeEffect_ConstIterable.*",
     // Linux LLVM9 c++17
     "std::_Rb_tree.*",
+    // archlinux
```
chrisduerr, line 484: Just putting it out there that there's a good chance this will affect other distros too in the future, Archlinux is likely just ahead of the curve. I'm on clang/llvm 20.1.7 if that says anything.

PR #1174, opened 2025-07-13
avoid deprecated

axkibe:
while I have not really an idea what I am doing, this seems to work and avoid deprecated switches.. 

PS: ```use gl_rs as gl;```
I don't know what gl_rs is, I always replaced it with simply ```use gl;```
Not changed in this PR.


pragmatrix:
> I don't know what gl_rs is,

The package got renamed to gl_rs in the Cargo.toml, most likely because of a conflict with the feature "gl". I am not sure if this is needed anymore, but we can leave it for now.

```
gl-rs = { version = "0.14.0", package = "gl" }
```

PR #1178, opened 2025-07-28
Fix `Rect::from_ltrb` parameter names

Nerixyz:
The parameter names in `Rect::from_ltrb` for right and bottom were swapped (but still correctly forwarded). When using inlay hints in a code editor, this was confusing. This PR swaps the names - nothing functional changes.


pragmatrix:
Thank you.

PR #1218, opened 2025-10-29
Update metal dependencies

pragmatrix:


nicoburns:
You may want to use `objc2_metal` too. We have that running here https://github.com/DioxusLabs/anyrender/blob/main/crates/anyrender_skia/src/metal.rs

PR #1224, opened 2025-11-17
Expose a `custom_empty` `FontMgr`

warrd:
Currently the only way to construct a `FontMgr` is with `new` or `empty`. The standard one scans all the system fonts, whereas the empty manager is useless in cases you actually want to render text, as you can't actually load any typefaces with it.

Here we forward to `SkFontMgr_New_Custom_Empty` (when freetype is available). This allows loading your own typefaces by calling `mgr.new_from_data`, without all the scanning of system fonts required at start up with the default manager.

As an example, on my laptop not scanning system fonts on my app cuts start up time from ~750ms to ~80ms.


pragmatrix:
Because users can instantiate a system font manager in response of a missing custom_empty() one, I think it's better to just return None if FreeType is not available.

Issue #861, opened 2023-10-22
issue: (probably not caused by skia-safe) Segmentation fault while closing window


marc2332:
Original issue https://github.com/marc2332/freya/issues/347

Hey Armin, do you know what could have caused that crash? I suppose you can't really do much to avoid it but maybe you know something I don't 

Thanks! 😄

pragmatrix:
As Simon already noted, drop order could be an issue. We've had a similar issue on Windows a while ago: https://github.com/rust-skia/rust-skia/issues/476#issuecomment-856058912

PR #1235, opened 2025-12-15
fix(gl-window-example): Implement proper drop of resources and abandon gr context in `gl-window` example

marc2332:
Fixes https://github.com/rust-skia/rust-skia/issues/861
Fixes https://github.com/marc2332/freya/issues/347

Today I got a new AMD gpu and my freya examples were crashing when I closed the windows, this fixes it

Does this make sense?


pragmatrix:
Thank you Marc, I've added a comment above this line so that it's clear why it's there and merge this without further investigation.

PR #1249, opened 2026-02-09
Port `metal-window` example to `objc2`. Add iOS support to example.

nicoburns:
Ports the `metal-window` example to use the `objc2` family of crates instead of `objc`, `cocoa`, `metal-rs`, `core-graphics-types`, and `foreign-types-shared` crates.

While I was at it I added support for building on iOS (I haven't actually run this example on iOS, but it is based on [code from anyrender_skia](https://github.com/DioxusLabs/anyrender/blob/main/crates/anyrender_skia/src/metal.rs) that is able to run apps on iOS).

### DIFF
--- a/skia-safe/Cargo.toml
+++ b/skia-safe/Cargo.toml
@@ -98,13 +98,20 @@ ash = { version = "^0.38.0" }
 vulkano = { version = "0.35" }
 
 # metal-window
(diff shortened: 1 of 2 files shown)


pragmatrix:
Looks good. Thank you.

PR #1283, opened 2026-04-29
feat: AVIF support

marc2332:
We would need https://github.com/marc2332/skia/commit/ea9571457452fcad1b1e28187c1745cd5ac6564a in your skia fork before merging this.

So, ignore the git submodule changing for now.

Full disclaimer: Claude Code basically did this, the code changes dont seem very complicated either. I tested it through Freya by building skia locally and it worked fine :)


pragmatrix:
Hey Marc,

what's the rationale for this? Why does this image format need to be supported in Skia, and isn't there a Rust alternative which could be used together with Skia's Bitmap or Pixmap types? Without these questions answered I don't see a justification for adding three repositories to the DEPS file that need to be downloaded for every fresh build. I hope you understand.

marc2332:
> Hey Marc,
> 
> what's the rationale for this? Why does this image format need to be supported in Skia, and isn't there a Rust alternative which could be used together with Skia's Bitmap or Pixmap types? Without these questions answered I don't see a justification for adding three repositories to the DEPS file that need to be downloaded for every fresh build. I hope you understand.

I mean, one could definitely use a rust alternative but same thing applies to e.g PNG, and that is supported, so I figured allowing AVIF would also make sense?

I didnt think those 3 new repos in the DEPS file would be a big deal

If you really dont want this, its fine, I might just leave it in my soft fork

Feel free to close the PR :v:

Issue #513, opened 2021-05-05
skia_safe::gpu::vk::BackendContext::set_max_api_version() should be an argument to skia_safe::gpu::vk::BackendContext::new()


remexre:
If omitted, Skia appears to default to "whatever we want," which can cause *really* confusing behavior (in my case, "fails to create context iff the Vulkan validation layers are enabled") when targeting a specific Vulkan version.

Issue #1274, opened 2026-03-28
More fixes for vulkan window example


maxnest0x0:
I can't believe it's a coincidence, I've spent the last few days struggling to fix validation errors, and today I see new changes from #1273 that fix them all!

However, there are still a few issues that break the example on my machine (LLVMpipe + Wayland):
1. I suspect Skia is trying to load v1.4 functions while the device only supports v1.3, resulting in `DirectContext` creation error. I fixed this with `BackendContext::set_max_api_version`.
```
library.api_version() = 1.4.341
instance.api_version() = 1.3.281
device.physical_device().api_version() = 1.3.281
device.api_version() = 1.3.281
Vulkan: failed to resolve vkTransitionImageLayout
Vulkan: failed to resolve vkCopyMemoryToImage
```
2. The example takes the first format from `surface_formats`, assumes it to be `B8G8R8A8_UNORM`, and panics if that's not the case.
```
device.physical_device().surface_formats(&surface, Default::default()).unwrap() = [
    (
        B8G8R8A8_SRGB,
        SrgbNonLinear,
    ),
    (
        B8G8R8A8_UNORM,
        SrgbNonLinear,
    ),
    // ...
]
```
3. After minimizing or hiding the window, when redraw is requested but swapchain is not recreated, `ImageLayout::PRESENT_SRC_KHR` is passed to `surface_for_framebuffer`, resulting in `wrap_backend_render_target` failure. I think `ImageLayout::UNDEFINED` should always be passed to Skia `ImageInfo` instead.

maxnest0x0:
Thank you for fixing this. I tested the example on two other devices and found new validation errors.

This happens on each redraw (caused by window resizing) on LLVMpipe + Wayland:
```
Validation Error: [ UNASSIGNED-non-acquired-swapchain-image-used ] | MessageID = 0x1cecb534
(Warning - This VUID has now been reported 10 times, which is the duplicate_message_limit value, this will be the last time reporting it).
vkQueueSubmit(): pSubmits[0] performs a layout transition on presentable VkImage 0x4d000000004d, but VkSemaphore 0xe000000000e and VkFence 0x160000000016 signaled by image acquire operation were not waited on.
The Vulkan spec states: After acquiring a presentable image and before modifying it, the application must use a synchronization primitive to ensure that the presentation engine has finished reading from the image (https://docs.vulkan.org/refpages/latest/refpages/source/VkSwapchainKHR.html#_description)
Objects: 2
[...]

PR #1292, opened 2026-06-06
feat(vulkan): introduce BackendContextBuilder for Vulkan context creation

### LINKED ISSUE #1291: Do not initialize vulkan memory allocator before maximum API version could be set
Since https://github.com/rust-skia/rust-skia/commit/a1fc79b2b5f25d56e6ead9f20afe141c7fc0cd77 a memory allocator is always created when creating a vulkan context, before `fMaxAPIVersion`  can be set.

This breaks on platforms (e.g. rendering using wgpu vulkan) where API version 1.4 is available but extension `VK_EXT_host_image_copy` is not when initializing a vulkan instance with version 1.3.

The fix there was setting `fMaxAPIVersion` to Version 1.3.0.

pragmatrix:
…tion

Closes #1291 and #513.


pragmatrix:
Related (max API version in the vulkan-example): https://github.com/rust-skia/rust-skia/issues/1274#issuecomment-4186442225

PR #1302, opened 2026-06-29
Add APIs to access font data as Data. (Allows sharing of font memory - issue #1092)

### LINKED ISSUE #1092: Would it be possible to access the font data without copying.
In Neovide we have this issue https://github.com/neovide/neovide/issues/298. Basically using emojis use an unreasonable amount of memory. I'm not completely sure about the cause, but most likely it's because we can't use the file mapped memory directly. Instead a copy has to be made here https://github.com/rust-skia/rust-skia/blob/24dee32a277b6c7b535755392a193373c133ce3b/skia-safe/src/core/typeface.rs#L248-L259

And it's not just the copy, it's also the fact that the whole file has to be read, so the whole file becomes mapped, and the memory usage becomes double of the font size. Although that can be purged from the physical memory again if the OS needs it.

I'm wondering if it would be possible to expose the data more directly?

willhoit:
This allows sharing font memory and memory mapped files with Skia in the manner Skia expects it.

* Rename Typeface::to_font_data to Typeface::to_font_bytes for consistency with the rest of the API
* Add Typeface::to_existing_font_data to return a Data attempting to share the underlying bytes, if possible
* Rename Typeface::new_from_data to Typeface::new_from_bytes
* Add a new version of Typeface::new_from_data using a Data object
* Update the type of the true type collection index to u32

Resolves #1092.

----------------------------------------

This seemed like the cleanest way to fit this functionality into the overall system since Skia hides a lot of the specifics of the fonts behind the scenes and SkData already provides all of the details. Plus this is how Skia itself does it.

The existing Rust API didn't fit the naming conventions of the rest of the library where `to_bytes` is used to mean `Vec<u8>`, so I also cleaned up that up (which makes this a minor API *breaking* change - but it's just a rename) and made the size of the collection index u32 (per the specs).

Testing (Linux, x86_64):
* `build`
* `build --release`
* `build test`
* `build --examples`
* ran `skia-org` and examined output
* `find -name \*.rs -print0 | xargs -0 grep to_font_data`
* `find -name \*.rs -print0 | xargs -0 grep new_from_data`
* used APIs in an unreleased application that loads dozens of fonts and does shaping (also saw improvements in performance due to shared data between the shaping and rendering)
* `cargo fmt`, `cargo clippy`

In my own applications:
* Examined output of `heaptrack` before and after, confirming that a 30MB font file was shared successfully
* Examined process mapping and confirmed the file was memory mapped
* In a one-off test harness confirmed that the sharing worked the other direction too, creating several instances of a typeface opened via Rust's `memmap2` crate

### DIFF
--- a/skia-bindings/src/bindings.cpp
+++ b/skia-bindings/src/bindings.cpp
@@ -1764,6 +1764,10 @@ extern "C" SkStreamAsset* C_SkTypeface_openStream(const SkTypeface* self, int* t
     return self->openStream(ttcIndex).release();
 }
 
+extern "C" SkStreamAsset* C_SkTypeface_openExistingStream(const SkTypeface* self, int* ttcIndex) {
+    return self->openExistingStream(ttcIndex).release();
+}
+
 extern "C" void C_SkTypeface_getBounds(const SkTypeface* self, SkRect* uninitialized) {
     new (uninitialized) SkRect(self->getBounds());
 }
@@ -2838,6 +2842,10 @@ extern "C" size_t C_SkStream_read(SkStream* stream, void* buffer, size_t len) {
 
 extern "C" size_t C_SkStreamAsset_getLength(const SkStreamAsset* self) {
     return self->getLength();
+}
+
+extern "C" const SkData* C_SkStreamAsset_getData(const SkStreamAsset* self) {
+    return self->getData().release();
 }
 
 extern "C" void C_SkWStream_destruct(SkWStream* self) {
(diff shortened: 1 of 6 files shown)


willhoit:
Windows and Mac have different font managers - the tests assumed that they would be able to return existing font data as a stream, but that was an invalid assumption.  

skshaper falls back on using table based access if `openExistingStream` return null, so the same thing would apply here -use  `copy_table_data` (which already returns a Data, this likely sharing the table data if possible)  

If so then just the tests will be changed and I'll update the documentation to provide a bit more guidance to consumers of the API.

pragmatrix:
Thank you a lot Ryan.

PR #1254, opened 2026-02-14
Initial support for Graphite

marcpabst:
Adds support for Skia's new Graphite backend.

- [x] wrapping initial Graphite API's in skia-bindings
- [x] making them available in skia-safe
- [x] enable interop with Metal on iOS/MacOS
- [ ] support interop with Vulkan, DX12

Note: This PR introduces a new `graphite` feature and a new `graphite` module which is completely seperate from the existing `gpu` module/feature. 

Addresses #1250, #1243

PR #1304, opened 2026-07-12
Add Graphite backend support (Metal & Vulkan)

hiking90:
This PR adds bindings for Skia's next-generation **Graphite** GPU backend behind a
new `graphite` cargo feature, with Metal and Vulkan context factories.

## API surface

New `skia_safe::graphite` module wrapping `skgpu::graphite`:

- **Core types**: `Context`, `Recorder`, `Recording`, `ContextOptions`,
  `RecorderOptions`, `BackendTexture`, `TextureInfo`, `InsertRecordingInfo`,
  `SubmitInfo`, `InsertStatus`
- **Backend entry points** (namespaced like `gpu::mtl` / `gpu::vk`):
  - `graphite::mtl::make_context` (+ `BackendContext`, `make_backend_texture`)
  - `graphite::vk::make_context` (reuses `gpu::vk::BackendContext`)
- **Surfaces / images**: `graphite::surfaces::{render_target, wrap_backend_texture}`,
  `graphite::images::{wrap_texture, texture_from_image, subset_texture_from}`,
  plus `as_image` / `as_image_copy`
- **Canvas integration**: `Canvas::recorder()` returns a lifetime-bound
  `BorrowedRecorder` for Graphite-backed canvases
- **Pixel readback**: `Context::read_pixels` — a synchronous wrapper over
  `asyncRescaleAndReadPixels` (Graphite has no `Surface::read_pixels`), the
  supported path for screenshots / golden-image tests

The `graphite` feature is included in `all-linux` / `all-windows` / `all-macos`.

## Ownership & soundness notes

- `Context` / `Recorder` are handed out as `std::unique_ptr` and are **not**
  ref-counted, so they are modeled as `RefHandle` (drop = `delete`), not `RCHandle`.
- `Context`, `Recorder` are `!Send + !Sync` (thread-affine, matching Ganesh's
  `DirectContext` / `RecordingContext`); `Recording` is `Send + !Sync` to allow the
  designed cross-thread hand-off from `Recorder::snap` to `Context::insert_recording`.
- `Canvas::recorder()` returns a dedicated `BorrowedRecorder` view that exposes
  read-only `Deref` plus forwarded `&mut` methods. It deliberately does **not**
  hand out `&mut Recorder`: that would let safe code `mem::replace` the
  surface-owned recorder out and double-free it.
- `InsertRecordingInfo::new` takes `&mut Recording` (`insertRecording` mutates the
  recording through the stored pointer), and the native struct is heap-allocated
  via `new`/`delete` shims: it embeds an `InsertStatus` holding a `std::string`,
  and a libstdc++ SSO string points into itself, so the struct must never be
  moved by value on the Rust side.
- `InsertStatus` re-exports the native `skgpu::graphite::InsertStatus::V` enum
  (all six statuses) instead of collapsing it to a bool-like result.

## Examples

| Example | Backend | What it exercises |
|---|---|---|
| `graphite_offscreen` | Metal | headless context/recorder lifecycle, snap → insert → submit |
| `graphite_offscreen_vulkan` | Vulkan | same on Vulkan (ash), plus readback verification and an RSS leak-check loop (`[iters] [reuse]` args) |
| `graphite_readback` | Metal | `Context::read_pixels` with pixel-level assertions |
| `metal-window2` | Metal | windowed rendering: `CAMetalLayer` drawable wrapped via `BackendTexture` + `surfaces::wrap_backend_texture`, per-frame snap/insert/submit (objc2-based) |

## Testing

- **macOS (Apple Silicon, Metal)**: `cargo check` / `clippy --all-targets` clean on
  `graphite,metal,vulkan,gpu` (also with `textlayout,svg`); doc tests pass;
  `graphite_readback` verifies correct pixels; `as_image` × 200 refcount-stable.
- **Arch Linux (AMD Radeon 780M, RADV)**: `graphite_offscreen_vulkan` passes with
  readback pixel verification; 500-iteration runs in both fresh and `reuse` modes
  show flat RSS (no leaks).
- The unsoundness fixed along the way was reproduced/validated with
  `#![forbid(unsafe_code)]` compile tests (the old `BorrowedRecorder` exploit no
  longer compiles).

## Not included / future work

- Dawn backend (not built by rust-skia).
- Precompilation / pipeline-caching APIs (`skgpu::graphite::PrecompileContext`).
- A windowed Vulkan + Graphite example (the swapchain-wrapping path currently has
  no example; `metal-window2` covers the equivalent on Metal).


pragmatrix:
Given this attempt to support Graphite seems to be more stable than #1254 and fits in nicely without changing the other feature configurations, it's a great start. Nevertheless, I will change some parts after the merge: For once, I am not satisfied with the directory structure, because it does not seem to mirror Skia's, but don't have any ideas about that yet. This seems to be related to the `gpu` feature and its module scope. Furthermore, I will remove some trivial tests and convert some FFI types to `Handle<T>` if the manual implementation matches what `Handle<T>` provides already. See https://github.com/rust-skia/rust-skia/commit/44d44afbce3e9026c6ba87caf940dbdbc7d16511 for some ideas how this might look like.

And thank you a lot @hiking90 for pushing this forward.

PR #1305, opened 2026-07-15
Recover from stale cached binary downloads

tronical:
When OUT_DIR/.cache contains a stale skia-binaries archive, for example one restored from an outdated CI cache, resuming the download with curl -C - is rejected by the server with HTTP 416 (Requested Range Not Satisfiable) if the cached file is already complete or larger than the remote file. Older curl versions (7.81 on Ubuntu 22.04) then fail with exit code 22, newer ones (8.7 on macOS) exit successfully and hand the possibly corrupt cached file to the unpacker. Either way the binaries download was considered failed and the build fell back to a full Skia source build - and kept doing so on every subsequent build, since the cached file was never cleaned up.

Recover in download_and_install: when the download or the unpacking of the archive fails while a cached download file exists, delete the cached file and retry once from scratch. A cached file that is complete and valid continues to be reused.

### DIFF
--- a/skia-bindings/build_support/binary_cache/download.rs
+++ b/skia-bindings/build_support/binary_cache/download.rs
@@ -192,6 +192,23 @@ fn should_try_download_binaries(
 }
 
 fn download_and_install(url: impl AsRef<str>, output_directory: &Path) -> io::Result<()> {
+    let url = url.as_ref();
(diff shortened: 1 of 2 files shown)


pragmatrix:
Nothing to add here, thank you @tronical. Merging.

PR #1308, opened 2026-07-28
Implement AsRef<[u8]> for Data to complement Deref

willhoit:
The standard library documentation for Deref [recommends](https://doc.rust-lang.org/std/convert/trait.AsRef.html#generic-implementations) implementers of Deref consider implementing AsRef too. For external libraries that work with shared bytes (fonts, images, etc.) but are not coupled to Skia itself this can allow for easier integration with APIs expecting impl AsRef<[u8]>.

------

## Testing Done

* `cargo test`
* `cargo build --examples --features gl`
* `cargo fmt`
* `cargo clippy`
*  Tested using a local project + [harfrust](https://github.com/harfbuzz/harfrust) with the feature flag `experimental_font_api` which uses `read_fonts` - see [FontBlob](https://github.com/googlefonts/fontations/blob/main/read-fonts/src/model/font/blob.rs)

## Note

The `RCHandle` also implements `AsRef` in a generic manner but only for types implementing traits that `[u8]` would not, so this seems okay.  If for some reason this isn't desired, it's possible for consumers of the API to add another structure to hold the `Data` instance and implement `AsRef` on that struct - just a bit more cumbersome.

### DIFF
--- a/skia-safe/src/core/data.rs
+++ b/skia-safe/src/core/data.rs
@@ -34,6 +34,12 @@ impl Deref for Data {
     }
 }
 
+impl AsRef<[u8]> for Data {
+    fn as_ref(&self) -> &[u8] {
+        self.deref().as_ref()
+    }
+}
+
 impl PartialEq for Data {
     // Although there is an implementation in SkData for equality testing, we
     // prefer to stay on the Rust side for that.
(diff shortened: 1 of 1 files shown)


pragmatrix:
Thank you @willhoit.

PR #1332, opened 2026-09-18
Android: keep the API level on the target passed to clang and bindgen

hiking90:
Building for Android from source fails with recent NDKs (seen on r29):

```
third_party/externals/icu/source/common/umapfile.cpp:251: error: use of undeclared identifier 'posix_madvise'
```

`posix_madvise` has been in bionic since API 23 and we build for API 26.

## Cause

`GnArgsBuilder` adds `--target=aarch64-linux-android` (the Rust target, with no API level) to `extra_cflags`. The NDK wrapper `aarch64-linux-android26-clang++` passes its own `--target=aarch64-linux-android26` first, and clang uses the last `--target`. With no level in the target, clang does not define `__ANDROID_MIN_SDK_VERSION__`:

```
$ clang --target=aarch64-linux-android   -dM -E -x c /dev/null | grep ANDROID_MIN_SDK
(nothing)
$ clang --target=aarch64-linux-android26 -dM -E -x c /dev/null | grep ANDROID_MIN_SDK
#define __ANDROID_MIN_SDK_VERSION__ 26
```

Newer bionic headers hide declarations based on that macro (`android/versioning.h`):

```c
#define __BIONIC_AVAILABILITY_GUARD(api_level) (__ANDROID_MIN_SDK_VERSION__ >= (api_level))
```

When the macro is not defined, it evaluates to 0, so every API behind this guard is hidden. The `-D__ANDROID_API__=26` we already pass does not help: r27 still checked `__ANDROID_API__` (see android/ndk#2081), but the new guard checks `__ANDROID_MIN_SDK_VERSION__`. CI does not hit this because its image uses NDK r26d.

Reproducer (NDK r29):

```
$ printf '#include <sys/mman.h>\nint f(void *p) { return posix_madvise(p, 4096, POSIX_MADV_NORMAL); }\n' > t.c
$ aarch64-linux-android26-clang -D__ANDROID_API__=26 -c t.c                                  # ok
$ aarch64-linux-android26-clang -D__ANDROID_API__=26 --target=aarch64-linux-android -c t.c   # error: call to undeclared function 'posix_madvise'
$ aarch64-linux-android26-clang -D__ANDROID_API__=26 --target=aarch64-linux-android26 -c t.c # ok
```

## Fix

`android.rs` appends `API_LEVEL` (already used for `ndk_api` and `-D__ANDROID_API__`) to the target:

- Skia build: `GnArgsBuilder::target`.
- bindgen and the bindings' C++ sources: `BindgenArgsBuilder::override_target`. This must be an override, because `skia_bindgen` appends the plain target after the platform args.

If the target comes from a `--target=` in `CC` and already ends in a level, e.g. `android30`, it is used as is.

## Effect on CI and published binaries

The CI builds also compile without an API level today, for the same reason. They only work because r26d checks `__ANDROID_API__`. With this change they compile for API 26. On r26d the headers should expose the same declarations, because `__ANDROID_API__` is 26 either way. The API-level-dependent codegen defaults I know of (e.g. emulated TLS below API 29) are the same for level 0 and 26. I have not compared r26d binaries before and after this change.

With a level in the target, clang defines `__ANDROID_API__` itself, so `-D__ANDROID_API__=26` now produces a `macro redefined` warning when the bindings' C++ sources compile. cargo-ndk users already see this warning, because cargo-ndk puts a leveled target into `CXXFLAGS`. I kept the `-D`, since I haven't checked that every supported NDK defines `__ANDROID_API__` from the triple.

## Testing

- `aarch64-linux-android`, NDK r29, Skia built from source via cargo-ndk: the build succeeds (it fails on `posix_madvise` without this change), `args.gn` gets `--target=aarch64-linux-android26`, and an app built this way runs on an Android 15 emulator.
- With bindgen at the level-less target and at `aarch64-linux-android26`, the generated `bindings.rs` is byte-identical.
- The target helper was checked on `aarch64`/`x86_64`/`i686-linux-android`, `armv7-linux-androideabi`, and targets already ending in `26`/`30`.
- `cargo clippy -p skia-bindings` reports no warnings for the build script.
- Tested on a Skia m152 checkout. `android.rs` and the `CC` target parsing are unchanged on current master.

## AI Assistance Disclosure

The root-cause investigation, the code change, and this description were produced with AI assistance (Claude Code). I reviewed and take responsibility for the final content.


pragmatrix:
Thanks @hiking90 ! Merging.

